PR #4, opened 2019-03-27
Add docs

joehoyle:
Reindexing Content
Search Index Modification
Elasticsearch Specifications
Synonyms


load.php
```diff
@@ -2,5 +2,11 @@
...
 
-add_action( 'plugins_loaded', __NAMESPACE__ . '\\bootstrap' );
-add_filter( 'hm_platform_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+use function HM\Platform\register_module;
+
+add_action( 'hm-platform.modules.init', function () {
+	register_module( 'search', __DIR__, 'Search', [ 'enabled' => true ], function () {
+		add_action( 'plugins_loaded', __NAMESPACE__ . '\\bootstrap' );
+		add_filter( 'hm_platform_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+	});
```
roborourke:
```suggestion
	} );
```

docs/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Pages
+- Media
+- Custom Post Types (registered with `show_in_search` or `public`)
+- Post's Meta
+- Post's Terms
+- Post's Author
+
+When used in conjunction with the Media Rekognition feature, all images are processed for automatic keyword detection and stored in the search index too.
+
+All documents that are uploaded to the media library are also parsed and indexed. For example, if you upload a PDF file, the PDF content will be read and included in the search index. Searches for keywords and phrases that are included in the document will be then be included in search results.
```
roborourke: I don't think this is enabled by default even if the plugin is installed, see #6

docs/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Posts
+- Pages
+- Media
+- Custom Post Types (registered with `show_in_search` or `public`)
+- Post's Meta
+- Post's Terms
+- Post's Author
+
+When used in conjunction with the Media Rekognition feature, all images are processed for automatic keyword detection and stored in the search index too.
```
roborourke: I think we might need to add a filter to add the image keywords meta field & alt text to the index if they're not in there already. Would be worth double checking.

docs/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Content that is indexed in the search index by default:
+
+- Posts
+- Pages
+- Media
+- Custom Post Types (registered with `show_in_search` or `public`)
+- Post's Meta
+- Post's Terms
+- Post's Author
```
roborourke: As of the latest version and WP 5.1 users are indexed too with improvements to `WP_User_Query`.

docs/elasticsearch-specifications.md
```diff
@@ -0,0 +1,31 @@
...
+
+- 6.2
+- 6.0
+- 5.5
+- 5.3
+- 5.1
+- 2.3
+- 1.5
+
+It's recommended to use the most up-to-date version. The specific version you want to use should be communicated to the Human Made team when requesting Elasticsearch be enabled.
```
roborourke: Altis team? Platform team? Support team?

docs/README.md
joehoyle: Hmm when I tried to activate it locally, it was already activated, so pretty sure it was on by default.

docs/elasticsearch-specifications.md
joehoyle: Yeah not sure, I think we'll correct once we go ahead with the name

docs/README.md
roborourke: Ok cool. Might want to be able to disable as it's a resource hog - especially on local, it can cause chassis to stack up unless you feed it lots of rams.

docs/README.md
joehoyle: 👍 added

docs/README.md
joehoyle: Ok, added a module setting for that!

PR #5, opened 2019-03-27
Enable elasticpress in the admin

### LINKED ISSUE #2: Turn on in admin by default

joehoyle:
To do this we also have to index all post types and post statuses. This isn't a big problem, because ES queries on the front end already pass the correct "status" permissions to the ES query, so we can't get data leakage this way.

Fixes #2


roborourke:
There are some settings that are on the admin page we should make configurable too such as indexing documents, search autocomplete, faceted search (that one we can probably make a more generic system for like we wanted for unison). Can add issues for those.

PR #23, opened 2019-05-11
Setup ES index on install

### LINKED ISSUE #16: Index is not created, causing 404s
Out of the box, the site's index doesn't exist, causing Query Monitor to display errors for the HTTP request. This is likely due to us not running the activation hooks.

joehoyle:
Use WP CLI hooks to run an index setup from multisite-install is run

Fixes #16.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -262,7 +262,7 @@ function get_elasticsearch_url() : string {
 }
 
 /**
- * When WordPress is installed via WP-CLI, run the ElasticPress setup
+ * When WordPress is installed via WP-CLI, run the ElasticPress setup.
  */
 function setup_elasticpress_on_install() {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -255,3 +260,12 @@ function get_elasticsearch_url() : string {
 	$host = sprintf( '%s://%s:%d', ELASTICSEARCH_PORT === 443 ? 'https' : 'http', ELASTICSEARCH_HOST, ELASTICSEARCH_PORT );
 	return $host;
 }
+
+/**
+ * When WordPress is installed via WP-CLI, run the ElasticPress setup
```
roborourke: Full stops at the end of comments
```suggestion
 * When WordPress is installed via WP-CLI, run the ElasticPress setup.
```

PR #30, opened 2019-05-24
Fix coding standards

roborourke:


inc/namespace.php
```diff
@@ -61,6 +69,7 @@ function bootstrap() {
 }
 
 function on_http_request_args( $args, $url ) {
```
rmccue: Undocumented functions make @rmccue sad.

roborourke: That wasn't me!

PR #35, opened 2019-07-26
Upgrade to ElasticPress 3.4.2

joehoyle:
ElasticPress 3 has a few breaking changes. I've accounted for the internal changes, but the main issue is EP 3 introduces this concept of "indexables" (which is nice n all), so there's an index per "type" of document. This means the default post index has changed from `$site-name-$id` to `$site-name-$indexable-$id`, so it requires a re-index by default.

We might be able to filter the post index to use the original index name, so we don't have to run reindexes on upgrade of Altis.


roborourke:
@joehoyle do you know how different the mapping is or the index data? We could potentially use ElasticSearch's index aliases to aid with migration by looping over indexes, detecting the old pattern and aliasing to the new one.

joehoyle:
Hmm no I don't know, it might not be different. I wonder if we could just filter the index name to use the old name rather than an alias?

rmccue:
> I wonder if we could just filter the index name to use the old name rather than an alias?

IMO, an alias is probably better, so that we can deprecate the old name with an eye towards moving to the new index names eventually.

pdewouters:
We should probably wait for a fix here https://github.com/10up/ElasticPress/pull/1829

roborourke:
@pdewouters can we not change how we're implementing feature activation?

pdewouters:
@roborourke the error is triggered by the admin notices - do we need those?

roborourke:
@pdewouters not if we're hiding the admin anyway but I think some are useful, I remember writing some code to specifically allow some admin notices and not others here so it'd better to check which ones are causing the errors and filter them specifically:

https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L587

I'm curious why making the feature inactive would cause that method to return false though. Perhaps the admin notice is getting triggered before ElasticPress has registered the default features.

pdewouters:
If I comment this out, the error in the dashboard disappears https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L56

```
//	if ( ! defined( 'EP_IS_NETWORK' ) ) {
//		define( 'EP_IS_NETWORK', true );
//	}
```

roborourke:
@pdewouters that's a no go to comment that out. It has to be in network mode to allow for full network search and full network indexing.

We need to get to the bottom of what the error messages are and why they're happening so we can either correct those issues through code or some other configuration.

pdewouters:
@roborourke that wasn't my suggestion.
It seems that the EP plugin needs to be loaded earlier on `plugins_loaded` so that the Features can be registered correctly.

roborourke:
Update here, Paul has tested this with a clean local server install and by running the indexing commands from ElasticPress. I'm going to do a second round of testing to double check but should be good to go after that.

roborourke:
EP 3.4 adds some additional features we need to decide what to do with:

- elasticpress debug bar error - `ep_is_indexing()` function has been removed, need to update or remove the code using it
- custom search results - basic post type. #76 adds a "search config" admin page under settings, propose bringing that to top level and placing custom search results listing & add-new pages as sub menu items of that

PR #37, opened 2019-08-05
Don't index post meta by default

joehoyle:


roborourke:
What's the reason behind this change? I think we definitely want post meta indexed if we're overriding `WP_Query` everywhere to use elastic. Is there a include / exclude approach we could take instead?

rmccue:
No response from submitter (cough cough @joehoyle), closing.

PR #41, opened 2019-10-29
Remove admin dashboard notices about syncing

roborourke:
EP is only interacted with through the CLI on Altis so this notice is just annoying.


nathanielks:
💯

PR #47, opened 2019-12-18
Detect if ELASTICSEARCH_HOST is empty

nathanielks:
This will prevent the module from bootstrapping if `ELASTICSEARCH_HOST` is empty, such as in the case no ElasticSearch infrastructure has been provisioned for an environment.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -28,7 +28,7 @@ function bootstrap() {
  * Load and configure Elasticpress.
  */
 function load_elasticpress() {
-	if ( ! defined( 'ELASTICSEARCH_HOST' ) ) {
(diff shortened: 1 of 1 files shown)


joehoyle:
Should we backport @nathanielks?

roborourke:
@joehoyle i think we should, it’s caused problems already. Although I wonder if we should trigger a warning if it’s empty that says how to disable this module

roborourke:
Merging for now, in terms of reporting back re. whether ES is set up or not requires more internal discussion.

nathanielks:
> Although I wonder if we should trigger a warning if it’s empty that says how to disable this module

I can see that being useful or annoying, depending. I like that it would point us in the specific direction required for fixing it, but if it's not dismiss-able it could be annoying for an environment who's not ready for search yet.

PR #56, opened 2020-02-13
Add ElasticPress debug bar

### LINKED ISSUE #55: Add ElasticPress debug bar
Query Monitor supports debug bar addons and 10up have this extension https://github.com/10up/debug-bar-elasticpress

Seems like it'd be really useful as debugging & profiling ElasticSearch can be painful at the best of times.

Acceptance criteria:

- [x] Plugin added to exclude list in humanmade/altis-core
- [x] Package `10up/debug-bar-elasticpress` is required in this module's `composer.json` at a fixed version (latest available)
- [x] Plugin is loaded when `Altis\get_config()['modules']['dev-tools']['query-monitor']` is `true`

**UPDATE**
After a discussion and the fact that we need to fix a lot of small things. We've decided to fork the plugin and fix it ourselves.

- fork the repo
- update the name in composer.json to be under the humanmade namespace
- fix the assets URL code 
- update the CSS
- publish to packagist

dashaluna:
Fixes #55

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
 	],
 	"require": {
 		"php": ">=7.1",
-		"10up/elasticpress": "^2.8.1"
+		"10up/elasticpress": "^2.8.1",
+		"10up/debug-bar-elasticpress": "1.4"
 	},
 	"autoload" : {
 		"files": [
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"require": {
 		"php": ">=7.1",
 		"10up/elasticpress": "^2.8.1",
-		"humanmade/debug-bar-elasticpress": "1.4"
+		"humanmade/debug-bar-elasticpress": "1.5"
 	},
 	"autoload" : {
 		"files": [
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"require": {
 		"php": ">=7.1",
 		"10up/elasticpress": "^2.8.1",
-		"10up/debug-bar-elasticpress": "1.4"
+		"humanmade/debug-bar-elasticpress": "1.4"
 	},
 	"autoload" : {
 		"files": [
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"require": {
 		"php": ">=7.1",
 		"10up/elasticpress": "^2.8.1",
-		"humanmade/debug-bar-elasticpress": "1.5"
+		"humanmade/debug-bar-elasticpress": "1.5.0"
 	},
 	"autoload" : {
 		"files": [
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -21,8 +21,12 @@ use WP_Query;
 
 function bootstrap() {
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_elasticpress' );
-	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_debug_bar_elasticpress' );
 	add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+
+	// Load debug bar for ElasticPress if Query Monitor is enabled in the config.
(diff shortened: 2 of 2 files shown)


inc/namespace.php
```diff
@@ -22,6 +22,11 @@
 function bootstrap() {
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_elasticpress' );
 	add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+
+	// Load debug bar for ElasticPress if enabled in the config.
+	if ( get_config()['modules']['search']['enabled'] ?? false ) {
```
roborourke: I've been a silly - this should've been to check that `get_config()['modules']['dev-tools']['query-monitor']` is true - sorry about that. If this code is running we already know the search module is enabled.

dashaluna: ✅

roborourke: You removed the check altogether? I meant to change it to check that query monitor was enabled.

dashaluna: OMG, now I've been a silly and totally misunderstood it. 🤦‍♀️sorry!

dashaluna:
@roborourke ok should be fixed.

inc/namespace.php
```diff
@@ -22,6 +22,11 @@
 function bootstrap() {
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_elasticpress' );
 	add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+
+	// Load debug bar for ElasticPress if Query Monitor is enabled in the config.
+	if ( get_config()['modules']['dev-tools']['query-monitor'] ?? false ) {
+		add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_debug_bar_elasticpress', 0 );
```
roborourke: One last addition here is to add `define( 'WP_EP_DEBUG', true );` to make sure elasticpress queries get logged. Wrap it in a conditional check to make it's not already defined though

composer.json
```diff
@@ -11,7 +11,8 @@
 	],
 	"require": {
 		"php": ">=7.1",
-		"10up/elasticpress": "^2.8.1"
+		"10up/elasticpress": "^2.8.1",
+		"10up/debug-bar-elasticpress": "1.4"
```
roborourke: Gonna have to update this to our version on packagist once you've published 😬

dashaluna:
@roborourke after ALL that, it's ready for review again :)

dashaluna:
@roborourke I've got a question.
So we forked the repo and added some code to it, that's been merged into `master`. However, I'm requesting here tag `1.4` which means that our new changes won't be loaded in. So do we need to ask for `master` branch or release a new tag?

roborourke:
@dashaluna we should only ever require a tag or tag range in a dependency - composer has a concept of minimum stability, if you require stable releases then it refuses to install anything that points a git branch because branches can change - tags are fixed to a specific commit.

Can you tag and release `1.5.0` for the EP debug bar, then we can submit patch releases if needed too

inc/namespace.php
```diff
@@ -22,6 +22,16 @@
 function bootstrap() {
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_elasticpress' );
 	add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+
+	// Load debug bar for ElasticPress if Query Monitor is enabled in the config.
+	if ( get_config()['modules']['dev-tools']['query-monitor'] ?? false ) {
+
+		// Enable debugging for Elastic Press Debug Bar to display query logs.
+		if ( ! defined( 'WP_EP_DEBUG' ) || ! WP_EP_DEBUG ) {
```
roborourke: This could cause a warning / error. If a constant is defined already you can't define it again as that would defeat the purpose of constants :)

dashaluna:
@roborourke ok, this is ready for review :)

PR #63, opened 2020-04-06
Enable the composer configuration of search

### LINKED ISSUE #6: Make EP built in features configurable
Elasticpress provides the following features that can be toggled on/off, some of which have additional settings.

- woocommerce faceted search
- related posts (our related posts plugin uses this feature if available)
- protected content (eg. index everything) - _enabled by default_
- document indexing - _enabled by default_
- faceted search
- user search (WP 5.1+ - this might be for ElasticPress v3 so can ignore for now)
- search autocomplete

These can configured using hooks and making calls to `ep_activate_feature()` and `ep_update_feature()` to add any settings.

There's some prior work here for making the autosuggest safe (not hitting ES directly) and some other bits https://github.com/humanmade/hm-platform/blob/beta/includes/settings.php#L71

To find where to set these features based on config see the below function:

https://github.com/humanmade/altis-enhanced-search/blob/0e00f75103cd89bbe8c1e3858f3be0ec60374f4d/inc/namespace.php#L300-L315

---------------

Acceptance criteria:

- [ ] Autosuggest feature can be enabled via config, default to off
  - [ ] Autosuggest modifications from [this prior code are implemented](https://github.com/humanmade/hm-platform/blob/beta/includes/settings.php#L71)
- [ ] Facets feature can be enabled via config, default to off
  - [ ] Facets config can be an object with a key `match-type` that defaults to `all` (see default feature settings in EP codebase)
    - Note this will require filtering the `ep_feature_settings` site option
- [ ] Woocommerce feature can be enabled via config, default to off
- [ ] Related Posts feature can be enabled via config, default to off
- [ ] Above features and their config are documented
  - For related posts also document the `ep_find_related()` function and note that a widget is registered
  - For autosuggest also document the added endpoint from the modifications in the link and not that it will act on the default WP search form template
  - For facets also document the `match-type` config option, explaining that it will force the results to match any selected taxonomy term or all selected taxonomy terms

smccafferty:
Resolves #6 

Add the ability to enabled/disable search components; related-posts, facets, woocommerce, autosuggest, index-documents, protected-content.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -304,12 +304,12 @@ function get_elasticpress_indexable_post_types( array $types ) : array {
 function override_elasticpress_feature_activation( bool $is_active, array $settings, EP_Feature $feature ) {
 	$config = get_config()['modules']['search'];
 	$features_activated = [
-		'search'            => true,
-		'related_posts'     => (bool) $config['related-posts'] ?? false,
-		'documents'         => (bool) $config['index-documents'] ?? true,
-		'facets'            => $config['facets'] ?? false,
-		'woocommerce'       => (bool) $config['woocommerce'] ?? false,
-		'autosuggest'       => (bool) $config['autosuggest'] ?? false,
+		'search' => true,
+		'related_posts' => (bool) $config['related-posts'] ?? false,
+		'documents' => (bool) $config['index-documents'] ?? true,
+		'facets' => $config['facets'] ?? false,
+		'woocommerce' => (bool) $config['woocommerce'] ?? false,
+		'autosuggest' => (bool) $config['autosuggest'] ?? false,
 		'protected_content' => (bool) $config['protected-content'] ?? true,
 	];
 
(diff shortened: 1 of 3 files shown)


inc/namespace.php
```diff
@@ -300,10 +304,13 @@ function get_elasticpress_indexable_post_types( array $types ) : array {
...
-		'related_posts' => false,
-		'documents'     => $config['index-documents'],
-		'facets'        => false,
+		'search'            => true,
+		'related_posts'     => (bool) $config['related-posts'] ?? false,
+		'documents'         => (bool) $config['index-documents'] ?? true,
+		'facets'            => $config['facets'] ?? false,
+		'woocommerce'       => (bool) $config['woocommerce'] ?? false,
+		'autosuggest'       => (bool) $config['autosuggest'] ?? false,
+		'protected_content' => (bool) $config['protected-content'] ?? true,
```
rmccue, line 313: We shouldn't align these, don't know why it was in the original

load.php
```diff
@@ -7,9 +7,14 @@
...
-		'enabled' => true,
-		'index-documents' => true,
+		'enabled'            => true,
+		'index-documents'    => true,
+		'related-posts'      => false,
+		'facets'             => false,
+		'woocommerce'        => false,
+		'autosuggest'        => false,
 		'slowlog_thresholds' => true,
+		'protected-content'  => true,
```
rmccue, line 17: No need to align these.

PR #65, opened 2020-04-24
Setting HM_ENV_TYPE throws a fatal error#

### LINKED ISSUE #58: Setting HM_ENV_TYPE throws a fatal error
```
Fatal error: Uncaught Aws\Exception\CredentialsException: Error retrieving credentials from the instance profile metadata service. (Client error: `GET http://169.254.169.254/latest/meta-data/iam/security-credentials/` resulted in a `404 Not Found` response: 
```

This is when trying to set `HM_ENV_TYPE` as `tests` for customizing the Altis modules in the tets environment.

The fatal error is thrown because of [the conditon](https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L118) inside the function `on_http_request_args` does not return early and the actual error is thrown from the method `sign_wp_request`

This can also be reproduced locally as well by setting this constant.

------------

Acceptance criteria:

- [ ] The check for environment against `local` should be updated to check against the type `development`, `staging` and `production`
- [ ] Setting a custom `HM_ENV_TYPE` value does not throw a fatal error when making requests to ElasticSearch eg. when doing a search in the admin posts list

abhishek-kaushik:
Resolves #58 

Check if Environment is not development, staging or production.

This will fix current problem , but if `HM_ENV_TYPE` changes to something else e.g. `staging->preprod` , we will get fatal again.

May be we could check for valid `HM_ENV_REGION`, then make the early return.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -114,7 +114,6 @@ function on_http_request_args( $args, $url ) {
 		return $args;
 	}
 
-
 	if ( Altis\get_environment_type() === 'local' || ! in_array( Altis\get_environment_architecture(), [ 'ec2', 'ecs' ], true ) ) {
 		return $args;
 	}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -152,7 +152,7 @@ function sign_wp_request( array $args, string $url ) : array {
  * @return RequestInterface
  */
 function sign_psr7_request( RequestInterface $request ) : RequestInterface {
-	if ( get_environment_type() === 'local' ) {
+	if ( Altis\get_environment_type() === 'local' ) {
 		return $request;
 	}
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -2,6 +2,7 @@
 
 namespace Altis\Enhanced_Search;
 
+use Altis;
 use Aws\Credentials;
 use Aws\Credentials\CredentialProvider;
 use Aws\Signature\SignatureV4;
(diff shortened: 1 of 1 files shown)


rmccue:
I think `HM_ENV_TYPE` is meant to be one of a few set strings, not an arbitrary string; I'm not sure if `preprod` is supported there.

cc @joehoyle

joehoyle:
I think any kind of condition based on environment type will not do, as other env types might or might not want request signing for ElasticSearch. We need a way to detect if ElasticSearch should be AWS signed or not. At the moment that's done based off the local env, but I think it would be better to get a more correct check.

As I understand it, running `composer dev-tools phpunit` will set the env to `tests` is that right? Or is this the case of a custom env type being used?

As for how we can detect if ES is expecting requests to be signed, we also have the `$is_cloud` conditional which we use in altis-cloud like https://github.com/humanmade/altis-cloud/blob/master/load.php#L9. This might be a good function to export from the module, and only add request signing in this module when that condition is met. Better still, perhaps we should move the request signing to the Cloud module, as I think it is cloud-specific.

Nikschavan:
> As I understand it, running composer dev-tools phpunit will set the env to tests is that right? Or is this the case of a custom env type being used?

For #58 I was trying to set `HM_ENV_TYPE` for running PHPUnit tests only, This was not for the custom environment altogether.

abhishek-kaushik:
running `composer dev-tools phpunit` does not set environment type as `test`, in stead it is `local`.
going with `$is_cloud` check and shall I create another ticket for moving request signing to cloud module or we should do in the same ticket.

abhishek-kaushik:
@joehoyle , do you think it's better to have the signing module under `enhanced-search` as we don't want to sign it, if `enhanced-search` is disabled.

joehoyle:
I think this approach is fine for now

PR #72, opened 2020-05-21
Reduce the looseness of the default search

roborourke:
The default search query was matching any of the search terms with a fuzziness of 1 allowing for a long tail of results. In some cases search results may not be sorted by relevance and the quantity of results becomes a problem. This removes the default EP search query that performs loose matching on any matching search term so now the longer or more specific a search is the fewer results will be returned.

This also introduces dynamic fuzziness when matching against all the provided search terms depending on the length of the search query.

### DIFF
--- a/docs/search-configuration.md
+++ b/docs/search-configuration.md
@@ -28,19 +28,18 @@ By setting `mode` to "advanced" it is possible to write more complex search term
 - A word followed by `*` will match anything with that prefix
 - A word preceded by `-` will be excluded from the results
 
-For example the following search would return results with exact matches for "the law", or the words "judge" and "dred" with some degree of fuzziness and excluding the word "marvel".
+For example the following search would return results with exact matches for "the law", or the words "judge" and "dred" and excluding the word "marvel".
 
 ```
-"the law" OR (judge AND dred) -marvel
+"the law" OR (judge AND dredd) -marvel
 ```
 
 ## Strict Search
-By default searching will match all the provided search terms with some degree of fuzziness to allow for misspellings and so on.
+By default searching will match _all_ the provided search terms.
 
-This means the more specific a search term is the fewer results will be provided. For example
-"the quick brown fox" will be interpreted as "the AND quick AND brown AND fox".
+This means the more specific a search query is the fewer results will be provided. For example "the quick brown fox" will be interpreted as "the `AND` quick `AND` brown `AND` fox".
 
-Setting `strict` to `false` will change the behaviour to matching each individual term in the search query, however full phrase matches will still be scored more highly.
+Setting `strict` to `false` will change the behaviour to match each individual word as well as the full search query, with complete matches being scored more highly.
 
 ```json
 {
--- a/docs/search-configuration.md
+++ b/docs/search-configuration.md
@@ -66,7 +66,7 @@ By default the post title, excerpt, content, author name and taxonomy terms are 
 		"altis": {
 			"modules": {
 				"search": {
-					"field-boost" {
+					"field-boost": {
 						"post_title": 3,
 						"post_excerpt": 2,
 						"meta.custom_field": 2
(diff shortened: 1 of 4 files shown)


roborourke:
@rmccue @joehoyle so we could just remove the fuzzy single word matching query when we're sorting by anything other than relevance but I'm not sure that the long tail of search results is particularly valuable for anyone anyway.

joehoyle:
Hmm yeah changing the results set when changing order by seems maybe odd? You'd see a different amount of pagination when you change the order by which I don't think is very expected?

joehoyle:
@roborourke I wonder if we could set up some test data or something for this. I'd like to also get feedback from the PMs etc who have raised this issue to they can verify the results are closer to expectations.

roborourke:
@joehoyle client was happy with this update, @yumito has more info and I can provide relevant GH issues privately in Slack.

They'd also like to see the ability to use advanced search syntax like supporting quoted text so might be worth making that configurable as part of this PR. It just involves changing to using a `simple_query_string` query in ES.

roborourke:
After a bit more discussion I think for the v4 release we'll use this stricter matching as a default but add config to leave it as the default, in addition we can provide a way to use the more advanced `simple_query_string` engine with support for brackets, and, or, custom fuzziness and quoted strings.

Will update the PR shortly.

docs/search-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+- A word preceded by `-` will be excluded from the results
+
+For example the following search would return results with exact matches for "the law", or the words "judge" and "dred" with some degree of fuzziness and excluding the word "marvel".
+
+```
+"the law" OR (judge AND dred) -marvel
+```
+
+## Strict Search
+By default searching will match all the provided search terms with some degree of fuzziness to allow for misspellings and so on.
```
joehoyle, line 38: I'm assuming `strict` is true by default, in which case why do we say "By default searching will match all the provided search terms with some degree of fuzziness to allow for misspellings and so on" which sounds like the opposite of strict?

docs/search-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+
+This means the more specific a search term is the fewer results will be provided. For example
```
joehoyle, line 40: I don't see how the above paragraph "means" this conclusion?

docs/search-configuration.md
roborourke: Yeah it is. Hmm, maybe we could flip it to “loose”. It’s essentially the difference between “or” and “and” so we could change it to “default_operator” instead

roborourke: I just thought a Boolean was simpler than a text value

joehoyle: ah ok, I read this paragraph describing not-strict (fuzz), but you are saying "By default search is strict (using AND operators), but does allow for fuzzy word matching still" is that right?

roborourke: Yeah that's correct

roborourke: I updated this to remove the references to fuzzy matching as it's needlessly confusing. I think that's ok as behind the scenes behaviour.

docs/search-configuration.md
```diff
@@ -0,0 +1,81 @@
...
+## Field Boosting
+By default the post title, excerpt, content, author name and taxonomy terms are searched against. Field boosting allows you to modify the importance of those fields with regards to search results. The following configuration will increase the important of the post title, excerpt and a custom meta field above the default value of 1.
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"search": {
+					"field-boost" {
```
joehoyle, line 69: Missing : here

rmccue, line 69:
```suggestion
					"field-boost": {
```

PR #73, opened 2020-06-05
Fix all linter errors

roborourke:


inc/namespace.php
```diff
@@ -107,11 +116,18 @@ function load_elasticpress() {
...
-	// @codingStandardsIgnoreLine
+/**
+ * Process HTTP request arguments.
+ *
+ * @param array $args Request arguments.
+ * @param string $url Request URL.
+ * @return array
+ */
+function on_http_request_args( array $args, string $url ) : array {
+	// phpcs:ignore
```
rmccue, line 130: Should ignore just the relevant rule.

inc/namespace.php
```diff
@@ -295,12 +325,12 @@ function get_elasticpress_indexable_post_types( array $types ) : array {
...
- * @param array $settings
- * @param EP_Feature $feature
- * @return void
+ * @param boolean $is_active True if the feature is active.
+ * @param array $settings Feature settings array.
+ * @param EP_Feature $feature The feature object.
+ * @return bool
  */
-function override_elasticpress_feature_activation( bool $is_active, array $settings, EP_Feature $feature ) {
+function override_elasticpress_feature_activation( bool $is_active, array $settings, EP_Feature $feature ) : bool {
```
rmccue, line 333: If `$feature->slug` is `facets`, this won't be `bool` right?

load.php
```diff
@@ -1,9 +1,13 @@
 <?php
+/**
+ * Altis Search Module.
+ *
+ * @package altis/search
+ */
 
-// phpcs:ignore
-namespace Altis\Enhanced_Search;
+namespace Altis\Enhanced_Search; // phpcs:ignore
```
rmccue, line 8: Should only ignore relevant rule.

inc/namespace.php
roborourke: Ah I see. So the function should always return a boolean, we don't actually want to return the potential array value of the facets config in this instance. I updated the code to reflect this more consistently.

load.php
roborourke: Actually not sure which rule it is but I think it'll disappear anyway once namespaces are allowed in `load.php`

PR #74, opened 2020-06-09
Add multilingual analyzers

roborourke:
This adds proper support for unicode characters as well as making use of the builr in stemmers and stop words for all supported languages.

Notable additions are:

- smartcn analyser for chinese languages
- proper use of kuromoji plugin with basic stop tags and lowercasing for roman characters
- ukrainian support
- polish support
- irish support
- brazilian portuguese support
- thai support
- bengali support

This update also ensure the post title is analysed using the site appropriate language and removes a deprecated feature using `_all` as a mapping field. This was unused.

Related https://github.com/humanmade/platform-dev/issues/373

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,4 +7,4 @@
 
 # Enhanced Search
 
-A better default search experience for Altis. Building off the [ElasticPress](https://github.com/10up/ElasticPress) plugin.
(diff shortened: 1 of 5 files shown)


roborourke:
@rmccue yeah it was already aligned more or less.

I think in a future update I'll expand those out to the full language name and update the JSON to reflect that, will be easier to find things than the 2 letter codes that way.

PR #75, opened 2020-06-09
Rename package to altis/search

### LINKED ISSUE #48: Disabled enhanced-search does not disable ElasticPress
When `enhanced-search` is disabled at a module (or environment) level, ElasticPress is _not_ actually disabled and displays a not configured warning in the dashboard if an ElasticSearch server has not been configured.

A quick look at this module suggests that this is because the module is being registered as `search` rather than `enhanced-search`: https://github.com/humanmade/altis-enhanced-search/blob/master/load.php#L14

Workaround is to define `altis.modules.search.enabled` instead of `altis.modules.enhanced-search.enabled`.

roborourke:
There has been a discrepancy between the registered module name and the package name since v1 that has caused some confusion. This will rectify that moving forward.

The package altis/enhanced-search will remain available on packagist but will not receive any updates.

Fixes #48

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-	"name": "altis/enhanced-search",
+	"name": "altis/search",
 	"description": "Enhanced Search module for Altis",
 	"type": "library",
 	"license": "GPL-2.0-or-later",


joehoyle:
mmm given the age and disruption to doing this, I am going to sale #shiphassailed on this one

rmccue:
We also no longer ship Elasticsearch as standard, so `enhanced` is actually a better name now anyway.

Issue #32, opened 2019-06-25
Support for Japanese


joehoyle:
In https://github.com/humanmade/platform-dev/issues/373 around Slack Chat https://hmn.slack.com/archives/C3CKS8LLR/p1560848386374400 we have been investigating support for Japanese.

To anyone like me who had no idea how Japanese "works" here's a run down of the situation.

With the current ElasticPress config, we are setting the [EP language](https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L356) to `cjk`. First problem: this breaks ElasticSearch, because the `ep_analyzer_language` filter is used for both the [analyzer](https://github.com/10up/ElasticPress/blob/b3d96a72dd54db1def357d6c0f093a2a182b6e06/includes/mappings/post/pre-5-0.php#L20) and in [a filter](https://github.com/10up/ElasticPress/blob/b3d96a72dd54db1def357d6c0f093a2a182b6e06/includes/mappings/post/pre-5-0.php#L45). Because `cjk` doesn't provide a stemmer, this breaks queries.

Anyway, `cjk` is only part of the issue. The changes needed to the index setting is much larger than just changing the default tokenizer's language.

So, a primer on Japanese as to why this doesn't work out of the box with ElasticPress: Japanese, for one, doesn't have a standard word separator like the SPACE in latin languages. This means off the bat there's basically no useful tokenizing of words. Also, Japanese has 4 charsets / alphabets (sorry for the lay understanding to anyone who speaks Japanese!), so there's no awareness of this from Elasticsearch out of the box.

To contend with the large language differences, there's as Elasticsearch plugin called [Kuromoji](https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-kuromoji.html) which we can use. This provides better analysers and tokenizers for Japanese. To use those though, we basically need to roll out own index settings for Japanese instead of the [EP index file](https://github.com/10up/ElasticPress/blob/b3d96a72dd54db1def357d6c0f093a2a182b6e06/includes/mappings/post/pre-5-0.php#L45).

Using an index file more along the lines of :

```
"analysis" => [
			"filter" => [
			  "kuromoji_rf" => [
				"type" => "kuromoji_readingform",
				"use_romaji" => "false"
			  ]
			],
			"tokenizer" => [
			  "kuromoji" => [
				"type" => "kuromoji_tokenizer"
			  ]
			],
			"analyzer" => [
			  "default" => [
				"type" => "custom",
[...]

PR #76, opened 2020-06-10
ElasticSearch user dictionary support

roborourke:
Adds a network admin and site admin settings page to upload or manually enter a user dictionary, stop words and synonyms.

Still to do:

- [x] Test with local environments, docker & chassis
- [x] Test AWS Elasticsearch Service integration
- [x] Filter ES mapping to:
  - Add user dictionary to kuromoji tokenizer
  - Add stop words filters for manual & uploaded to default analyzer
  - Add synonyms filters for manual & uploaded to default analyzer
  - Support network wide override for sites with same language
- [x] Handle reindexing on update using `_update_by_query` ES endpoint

See #32.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -774,18 +774,20 @@ function elasticpress_mapping( array $mapping ) : array {
 				continue;
 			}
 
-			if ( $type === 'synonyms' ) {
-				$synonyms[ "{$sub_type}_{$type}_filter" ] = [
-					'type' => 'synonym_graph',
-					'synonyms_path' => $package_id,
-				];
-			}
-			if ( $type === 'stopwords' ) {
-				$stopwords[ "{$sub_type}_{$type}_filter" ] = [
-					'type' => 'stop',
-					'ignore_case' => true,
-					'stopwords_path' => $package_id,
-				];
+			switch ( $type ) {
+				case 'synonyms':
+					$synonyms[ "{$sub_type}_{$type}_filter" ] = [
+						'type' => 'synonym_graph',
+						'synonyms_path' => $package_id,
+					];
+					break;
+				case 'stopwords':
+					$stopwords[ "{$sub_type}_{$type}_filter" ] = [
+						'type' => 'stop',
+						'ignore_case' => true,
+						'stopwords_path' => $package_id,
+					];
+					break;
 			}
 		}
 	}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -578,7 +578,9 @@ function elasticpress_mapping( array $mapping ) : array {
 		unset( $mapping['settings']['mappings']['post']['properties']['post_title']['fields']['post_title']['analyzer'] );
 	}
 
-	// Handle user dictionary for Japanese tokenizer.
+	/**
+	 * Handle user dictionary for Japanese sites.
+	 */
 	if ( $language === 'ja' ) {
 		$is_network_japanese = get_site_option( 'WPLANG', 'en_US' ) === 'ja';
 		$user_dictionary_file = Packages\get_package_file_path( 'uploaded-user-dictionary' );
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -130,7 +130,7 @@ function load_elasticpress() {
 	add_filter( 'register_post_type_args', __NAMESPACE__ . '\\custom_search_results_post_type_args', 10, 2 );
 
 	// Set up packages feature.
-	Packages\setup();
+	Packages\bootstrap();
 }
 
 /**
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -862,11 +862,11 @@ function custom_search_results_post_type_args( array $args, string $post_type ) 
 		return $args;
 	}
 
-	// Use the built in search icon.
-	$args['menu_icon'] = 'dashicons-search';
+	// Hide in admin menu, we'll add it as a subitem of main search config page.
+	$args['show_in_menu'] = 'search-config';
 
 	// Change the menu name to something shorter.
-	$args['labels']['menu_name'] = _x( 'Search Config', 'post type menu name', 'altis' );
+	$args['labels']['all_items'] = _x( 'Custom Search Results', 'post type menu name', 'altis' );
 
 	return $args;
 }
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -101,6 +101,9 @@ function load_elasticpress() {
 	// Filter Options for Facet component settings.
 	add_filter( 'site_option_ep_feature_settings', __NAMESPACE__ . '\\filter_facet_settings' );
 	add_filter( 'option_ep_feature_settings', __NAMESPACE__ . '\\filter_facet_settings' );
+
+	// Set up packages feature.
+	Packages\setup();
 }
 
 /**
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -585,15 +585,14 @@ function elasticpress_mapping( array $mapping ) : array {
 	 */
 	if ( $language === 'ja' ) {
 		$is_network_japanese = get_site_option( 'WPLANG', 'en_US' ) === 'ja';
-		$user_dictionary_file = Packages\get_package_file_path( 'uploaded-user-dictionary' );
-		if ( ! file_exists( $user_dictionary_file ) && $is_network_japanese ) {
-			$user_dictionary_file = Packages\get_package_file_path( 'global-uploaded-user-dictionary' );
+		$user_dictionary_package_id = Packages\get_package_id( 'uploaded-user-dictionary' );
+		if ( ! $user_dictionary_package_id && $is_network_japanese ) {
+			$user_dictionary_package_id = Packages\get_package_id( 'global-uploaded-user-dictionary' );
 		}
 
 		// Check for a package ID and add it to the kuromoji tokenizer.
-		$package_id = Packages\get_package_id( $user_dictionary_file );
-		if ( $package_id ) {
-			$mapping['settings']['analysis']['tokenizer']['kuromoji']['user_dictionary'] = $package_id;
+		if ( $user_dictionary_package_id ) {
+			$mapping['settings']['analysis']['tokenizer']['kuromoji']['user_dictionary'] = $user_dictionary_package_id;
 		}
 	}
 
@@ -603,21 +602,20 @@ function elasticpress_mapping( array $mapping ) : array {
 	 * Synonyms and stopwords are quick enough to be applied at search time and avoid
 	 * increasing the index size unnecessarily.
 	 */
-	$types = [ 'synonyms', 'stopwords' ];
 	$is_network_language = get_site_option( 'WPLANG', 'en_US' ) === get_option( 'WPLANG', 'en_US' );
 	$synonyms = [];
 	$stopwords = [];
 
-	foreach ( $types as $type ) {
+	foreach ( [ 'synonyms', 'stopwords' ] as $type ) {
 		foreach ( [ 'uploaded', 'manual' ] as $sub_type ) {
 			// Get package file path.
-			$file = Packages\get_package_file_path( "{$sub_type}-{$type}" );
-			if ( ! file_exists( $file ) && $is_network_language ) {
-				$file = Packages\get_package_file_path( "global-{$sub_type}-{$type}" );
+			$package_id = Packages\get_package_id( "{$sub_type}-{$type}" );
+			// Check for network default.
+			if ( ! $package_id && $is_network_language ) {
+				$package_id = Packages\get_package_id( "global-{$sub_type}-{$type}" );
 			}
 
-			// Look for a package ID.
-			$package_id = Packages\get_package_id( $file );
+			// Check for a package ID.
 			if ( ! $package_id ) {
 				continue;
 			}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -587,7 +587,7 @@ function elasticpress_mapping( array $mapping ) : array {
 		$is_network_japanese = get_site_option( 'WPLANG', 'en_US' ) === 'ja';
 		$user_dictionary_package_id = Packages\get_package_id( 'uploaded-user-dictionary' );
 		if ( ! $user_dictionary_package_id && $is_network_japanese ) {
-			$user_dictionary_package_id = Packages\get_package_id( 'global-uploaded-user-dictionary' );
+			$user_dictionary_package_id = Packages\get_package_id( 'uploaded-user-dictionary', true );
 		}
 
 		// Check for a package ID and add it to the kuromoji tokenizer.
@@ -612,7 +612,7 @@ function elasticpress_mapping( array $mapping ) : array {
 			$package_id = Packages\get_package_id( "{$sub_type}-{$type}" );
 			// Check for network default.
 			if ( ! $package_id && $is_network_language ) {
-				$package_id = Packages\get_package_id( "global-{$sub_type}-{$type}" );
+				$package_id = Packages\get_package_id( "{$sub_type}-{$type}", true );
 			}
 
 			// Check for a package ID.
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -395,10 +395,11 @@ function elasticpress_analyzer_language() : string {
 		'ar'             => 'ar', // arabic.
 		'hy'             => 'hy', // armenian.
 		'eu'             => 'eu', // basque.
-		'pt_br'          => 'pt', // brazilian portuguese.
+		'pt_br'          => 'br', // brazilian portuguese.
 		'bg_bg'          => 'bg', // bulgarian.
+		'bn_bd'          => 'bn', // bengali.
 		'ca'             => 'ca', // catalan.
-		'cs_cz'          => 'cs', // czeh.
+		'cs_cz'          => 'cs', // czech.
 		'da_dk'          => 'da', // danish.
 		'nl_be'          => 'nl', // dutch.
 		'nl_nl'          => 'nl',
@@ -413,6 +414,7 @@ function elasticpress_analyzer_language() : string {
 		'fr_be'          => 'fr', // french.
 		'fr_ca'          => 'fr',
 		'fr_fr'          => 'fr',
+		'ga'             => 'ga', // irish.
 		'gl_es'          => 'gl', // galician.
 		'de_at'          => 'de', // german.
 		'de_ch'          => 'de',
@@ -426,8 +428,8 @@ function elasticpress_analyzer_language() : string {
 		'it_it'          => 'it', // italian.
 		'lv'             => 'lv', // latvian.
 		'lt_lt'          => 'lt', // lithuanian.
-		'nb_no'          => 'no', // norwegian.
-		'nn_no'          => 'no',
+		'nb_no'          => 'nb', // norwegian bokmål.
+		'nn_no'          => 'nn', // norwegian nynorsk.
 		'fa_ir'          => 'fa', // persian.
 		'pl_pl'          => 'pl', // polish.
 		'pt_pt'          => 'pt', // portuguese.
@@ -516,7 +518,7 @@ function elasticpress_mapping( array $mapping ) : array {
 	$mapping['settings']['analysis']['analyzer']['shingle_analyzer'] = [
 		'type' => 'custom',
 		'tokenizer' => 'icu_tokenizer',
-		'filter' => [ 'icu_normalizer', 'icu_folding', 'lowercase', 'shingle_filter' ],
+		'filter' => [ 'icu_normalizer', 'icu_folding', 'shingle_filter' ],
 	];
 
 	// Get analyzer language.
@@ -529,10 +531,6 @@ function elasticpress_mapping( array $mapping ) : array {
 			$mapping['settings']['analysis']['analyzer']['default']['char_filter'] ?? [],
 			[ 'html_strip' ]
 		);
-		// If the last filter was icu_folding then add lowercase filter.
-		if ( end( $mapping['settings']['analysis']['analyzer']['default']['filter'] ) === 'icu_folding' ) {
-			$mapping['settings']['analysis']['analyzer']['default']['filter'][] = 'lowercase';
-		}
 	}
 
 	// Remove deprecated _all parameter.
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -8,6 +8,7 @@
 namespace Altis\Enhanced_Search;
 
 use Altis;
+use Altis\Enhanced_Search\Packages;
 use Aws\Credentials;
 use Aws\Credentials\CredentialProvider;
 use Aws\Signature\SignatureV4;
@@ -15,7 +16,6 @@ use ElasticPress_CLI_Command;
 use EP_Dashboard;
 use EP_Feature;
 use EP_Features;
-
 use GuzzleHttp\Psr7\Request;
 use Psr\Http\Message\RequestInterface;
 use WP_CLI;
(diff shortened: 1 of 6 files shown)


inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+		$file_date_var = "{$type}_file_date";
+		$uploaded_file_status = "{$type}_uploaded_status";
+		$manual_file_status = "{$type}_manual_status";
+		$uploaded_file_error = "{$type}_uploaded_error";
+		$manual_file_error = "{$type}_manual_error";
+
+		$file_type_string = str_replace( '_', '-', $type );
+
+		if ( $for_network ) {
+			$$uploaded_file_status = get_site_option( "altis_search_package_status_uploaded-{$file_type_string}" );
```
rmccue, line 98: Are these intentionally variable variables? Can we use an array or something instead to make this more explicit? Variable variables are very easy to miss while reading the code.

(If it's for ergonomics for the template file, then an `extract()` of the data later would be better, but the arrays are probably fine anyway.)

Sidenote, I don't love how variable these option names are, and they might also start getting close to the maximum option name length?

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+			$$file_date_var = filemtime( $$uploaded_file_var );
+		}
+	}
+
+	$did_update = isset( $_GET['did_update'] );
+
+	if ( $for_network ) {
+		$errors = get_site_option( 'altis_search_config_error', false );
+	} else {
+		$errors = get_option( 'altis_search_config_error', false );
```
rmccue, line 129: These are probably good cases for transients if we can.

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+	}
+}
+
+/**
+ * Sets the search config error message.
+ *
+ * @param WP_Error $error The error object to add.
+ * @param bool $for_network Whether this should be added as a network level error.
+ */
+function add_error_message( WP_Error $error, bool $for_network = false ) {
```
rmccue, line 155:
```suggestion
function add_error_message( WP_Error $error, bool $for_network = false ) : void {
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+			delete_option( 'altis_search_config_error' );
+		}
+	}
+}
+
+/**
+ * Sets the search config error message.
+ *
+ * @param WP_Error $error The error object to add.
+ * @param bool $for_network Whether this should be added as a network level error.
```
rmccue, line 153:
```suggestion
 * @param bool $for_network Whether this should be added as a network level error.
 * @return void
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+		if ( $for_network ) {
+			delete_site_option( 'altis_search_config_error' );
+		} else {
+			delete_option( 'altis_search_config_error' );
+		}
+	}
+}
+
+/**
+ * Sets the search config error message.
```
rmccue, line 150:
```suggestion
 * Set the search config error message.
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+	 * @param string $path Absolute directory path, defaults to wp-content/es-packages.
+	 */
+	$path = apply_filters( 'altis.search.packages_dir', $path );
+
+	return $path;
+}
+
+/**
+ * Get a package file path. Returns the path if the file exists
+ * or null on failure.
```
rmccue, line 195: Short and long descriptions should be split. Also, the indicated return value is wrong anyway. Suggest removing second sentence entirely.

```suggestion
 * Get a package file path.
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+
+	return $package_id;
+}
+
+/**
+ * Process the packages form.
+ *
+ * @return void
+ */
+function handle_form() {
```
rmccue, line 243:
```suggestion
function handle_form() : void {
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+	array_map( __NAMESPACE__ . '\\add_error_message', $errors );
+
+	// Schedule mapping and index updates.
+	if ( empty( $errors ) ) {
+		do_action( 'altis.search.updated_packages', $for_network, $should_update_indexes );
+	}
+
+	// Redirect back to form.
+	$redirect_url = is_network_admin() ? network_admin_url( 'settings.php' ) : admin_url( 'admin.php' );
+	wp_safe_redirect( $redirect_url . '?page=search-config&did_update=1' );
```
rmccue, line 343: `add_query_arg()` would be better here.

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+ * Create and associate a package file with AES, removing any existing
+ * files with a matching name.
+ *
+ * @param string $slug The package slug.
+ * @param string $file The package file path.
+ * @param bool $for_network If true creates the package at the network level.
+ * @return string|WP_Error The package ID for referencing in analysers or a WP_Error.
+ */
+function create_package( string $slug, string $file, bool $for_network = false ) {
+
```
rmccue, line 370: Bonus newline.
```suggestion
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+			sprintf( __( 'Package for slug %s does not exist', 'altis' ), $slug )
+		);
+	}
+
+	$file = get_package_path( $slug, $for_network );
+
+	// Ensure file exists.
+	if ( ! file_exists( $file ) ) {
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		trigger_error( sprintf( 'The package file %s does not exist.', $file ), E_USER_WARNING );
```
rmccue, line 445: If we're just deleting the file, do we need to error? Seems a bit "failed successfully" to me

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+ *
+ * @param integer|null $blog_id The site to get index names for. Defaults to current site.
+ * @return string
+ */
+function get_site_indices( ?int $blog_id = null ) : string {
+	$indexes = [];
+
+	// Check for default indexes and their existence.
+	foreach ( Indexables::factory()->get_all() as $indexable ) {
+		if ( ! $indexable->index_exists( $blog_id ) ) {
```
rmccue, line 496: Can we default `$blog_id` ourselves to guard against potential inconsistent behaviour with null across these different APIs?

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+ * If the data needs to be updated such as if the Japanese user dictionary
+ * is updated then $update_index allows for updating using `_update_by_query`.
+ *
+ * Note this function should be run as a background task only.
+ *
+ * @param boolean $for_network Whether this is a network wide mapping update.
+ * @param boolean $update_data Whether the index data should be updated as well.
+ * @return void
+ */
+function do_settings_update( bool $for_network = false, bool $update_data = false ) {
```
rmccue, line 518:
```suggestion
function do_settings_update( bool $for_network = false, bool $update_data = false ) : void {
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,582 @@
...
+/**
+ * Update index settings.
+ *
+ * @param string $index An optional index name or pattern to operate on.
+ * @param array $settings The updated settings.
+ * @param boolean $update_data Whether to update data in the index.
+ * @return void
+ */
+function update_index_settings( string $index, array $settings, bool $update_data = false ) {
+
```
rmccue, line 550:
```suggestion
function update_index_settings( string $index, array $settings, bool $update_data = false ) : void {
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+<style>
+	.columns { display: flex; flex-wrap: wrap; }
+	.column { flex: 1 1 300px; width: 300px; margin-right: 20px; }
+	.column textarea { width: 100%; }
+	.column pre { background: #fff; border: 1px solid #152a4e; padding: 10px; overflow: auto; }
+	.column ul li { margin-left: 20px; list-style: disc; }
+	.es-package-status { float: right; text-transform: uppercase; padding: 2px 5px; font-size: 65%; border-radius: 3px; background: #152a4e; color: #fff; }
+	.es-package-status--active { background-color: #3FCF8E; color: #000; }
+	.es-package-status[class*="error"] { background-color: #ED7B9D; color: #152a4e; }
+</style>
```
rmccue, line 18: Oof, can this be in a stylesheet instead?

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+					<pre><?php esc_html_e( 'text, token 1 ... token n, reading 1 ... reading n, part-of-speech tag', 'altis' ); ?></pre>
+					<p><?php esc_html_e( 'For example', 'altis' ); ?>:</p>
+					<pre>東京スカイツリー,東京 スカイツリー,トウキョウ スカイツリー,カスタム名詞</pre>
+				</div>
+				<div class="column">
+					<h3>
+						<label for="user-dictionary-file"><?php esc_html_e( 'File upload', 'altis' ); ?></label>
+						<?php if ( $user_dictionary_file_date && $user_dictionary_uploaded_status ) : ?>
+							<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $user_dictionary_uploaded_status ) ); ?>">
+								<?php esc_html_e( 'Status', 'altis' ); ?>: <?php echo esc_html( $user_dictionary_uploaded_status ); ?>
```
rmccue, line 59: This needs to be done as a single string, not appended manually, for languages with varying orders.

```suggestion
								<?php echo esc_html( sprintf( __( 'Status: %s', 'altis' ), $user_dictionary_uploaded_status ) ) ?>
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+			<div class="column">
+				<p><?php esc_html_e( 'Synonyms allow the search index to treat specific words or phrases as equal. This is useful if:', 'altis' ); ?></p>
+				<ul>
+					<li><?php esc_html_e( 'you have a lot of acronyms or other contractions in your content', 'altis' ); ?></li>
+					<li><?php esc_html_e( 'you have terms in the content that do not appear in the dictionary', 'altis' ); ?></li>
+					<li><?php esc_html_e( 'there are common mis-spellings of search terms', 'altis' ); ?></li>
+					<li><?php esc_html_e( 'you have users from different countries who may use different words for the same thing', 'altis' ); ?></li>
+				</ul>
+				<p><?php esc_html_e( 'The syntax for synonyms is as follows:', 'altis' ); ?></p>
+				<pre># <?php esc_html_e( 'Comments are allowed using a hash at the start the line.', 'altis' ); ?></pre>
```
rmccue, line 101:
```suggestion
				<pre><?php esc_html_e( '# Comments are allowed using a hash at the start the line.', 'altis' ); ?></pre>
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+				<pre>i-pod, i pod => ipod
+tent => bivouac, teepee
+sea biscuit, sea biscit => seabiscuit</pre>
+			</div>
+			<div class="column">
+				<h3>
+					<label for="synonyms-file"><?php esc_html_e( 'File upload', 'altis' ); ?></label>
+					<?php if ( $synonyms_file_date && $synonyms_uploaded_status ) : ?>
+						<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $synonyms_uploaded_status ) ); ?>">
+							<?php esc_html_e( 'Status', 'altis' ); ?>: <?php echo esc_html( $synonyms_uploaded_status ); ?>
```
rmccue, line 116:
```suggestion
							<?php esc_html( sprintf( __( 'Status: %s', 'altis' ), $synonyms_uploaded_status ) ) ?>
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+					<input type="submit" class="components-button is-link is-destructive" name="synonyms-remove" value="<?php esc_attr_e( 'Remove synonyms file', 'altis' ); ?>" />
+				<?php endif; ?>
+				<p>
+					<input type="file" accept="text/plain" id="synonyms-file" name="synonyms-file" />
+				</p>
+				<h3>
+					<label for="synonyms-text"><?php esc_html_e( 'Manual entry', 'altis' ); ?></label>
+					<?php if ( $synonyms_manual_status ) : ?>
+						<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $synonyms_manual_status ) ); ?>">
+							<?php esc_html_e( 'Status', 'altis' ); ?>: <?php echo esc_html( $synonyms_manual_status ); ?>
```
rmccue, line 145:
```suggestion
							<?php echo esc_html( sprintf( __( 'Status: %s', 'altis' ), $synonyms_manual_status ) ); ?>
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+				<pre>ignore
+me
+please</pre>
+			</div>
+			<div class="column">
+				<h3>
+					<label for="stopwords-file"><?php esc_html_e( 'File upload', 'altis' ); ?></label>
+					<?php if ( $stopwords_file_date && $stopwords_uploaded_status ) : ?>
+						<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $stopwords_uploaded_status ) ); ?>">
+							<?php esc_html_e( 'Status', 'altis' ); ?>: <?php echo esc_html( $stopwords_uploaded_status ); ?>
```
rmccue, line 175:
```suggestion
							<?php echo esc_html( sprintf( __( 'Status: %s', 'altis' ), $stopwords_uploaded_status ) ) ?>
```

inc/packages/templates/config.php
```diff
@@ -0,0 +1,222 @@
...
+					<input type="submit" class="components-button is-link is-destructive" name="stopwords-remove" value="<?php esc_attr_e( 'Remove stop words file', 'altis' ); ?>" />
+				<?php endif; ?>
+				<p>
+					<input type="file" accept="text/plain" id="stopwords-file" name="stopwords-file" />
+				</p>
+				<h3>
+					<label for="stopwords-text"><?php esc_html_e( 'Manual entry', 'altis' ); ?></label>
+					<?php if ( $stopwords_manual_status ) : ?>
+						<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $stopwords_manual_status ) ); ?>">
+							<?php esc_html_e( 'Status', 'altis' ); ?>: <?php echo esc_html( $stopwords_manual_status ); ?>
```
rmccue, line 204:
```suggestion
							<?php echo sprintf( esc_html( __( 'Status: %s', 'altis' ), $stopwords_manual_status ) ) ?>
```

rmccue (changes requested):

inc/namespace.php
```diff
@@ -614,6 +620,78 @@ function elasticpress_mapping( array $mapping ) : array {
 		unset( $mapping['mappings']['post']['properties']['post_title']['fields']['post_title']['analyzer'] );
 	}
 
+	/**
+	 * Handle user dictionary for Japanese sites.
+	 */
+	if ( $language === 'ja' ) {
```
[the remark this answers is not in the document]
roborourke: User dictionary is only ever used in the kuromoji analysis plugin so it's scoped to Japanese language sites specifically.

inc/namespace.php
```diff
@@ -614,6 +620,78 @@ function elasticpress_mapping( array $mapping ) : array {
...
+		$is_network_japanese = get_site_option( 'WPLANG', 'en_US' ) === 'ja';
```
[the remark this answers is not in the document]
roborourke: Network vs current site. Packages can be added network wide or overridden per site.

inc/packages/namespace.php
roborourke: Yeah I just threw the page together fairly quickly and it went through a lot of iteration in that time. I'm not really into using `extract()` so I think an array will be better as you say.

I did worry about max option name length but for the hard coded values currently in use we're ok.

inc/packages/templates/config.php
roborourke: It could be yeah, I just found this easier during development

inc/packages/templates/config.php
roborourke: I don't really get this change - the hash shouldn't be translated as it's to demonstrate the comment.

rmccue: In different languages it may need to be in different places in the sentence; e.g. in RTL languages, there may need to be LTR override to ensure it's displayed in the right place.

inc/namespace.php
```diff
@@ -614,6 +620,78 @@ function elasticpress_mapping( array $mapping ) : array {
...
+			if ( ! $package_id && $is_network_language ) {
+				$package_id = Packages\get_package_id( "{$sub_type}-{$type}", true );
+			}
+
+			// Check for a package ID.
+			if ( ! $package_id ) {
+				continue;
+			}
+
+			if ( $type === 'synonyms' ) {
```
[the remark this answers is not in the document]
roborourke: True, but less indentation this way and only 2 options. I'll update it though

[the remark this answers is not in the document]
rmccue: Yeah, makes it clearer that the options are mutually exclusive though imo which helps with legibility

inc/packages/namespace.php
roborourke: I don't totally follow. A lot of functions in WP work this way where null is implied to be the current "thing", especially in the multisite functions.

inc/packages/templates/config.php
roborourke: It's a code example. Is code written RTL in RTL languages?

rmccue: No, which is why there may need to be an LTR override as part of the translated text.

In RTL languages, the document's language may be set to RTL. In order to display this interleaved text directly, an [LTR-override character](https://www.fileformat.info/info/unicode/char/200e/index.htm) may be required.

(The examples not necessarily because it's not interleaved, otherwise I'd have suggested making those translatable too.)

roborourke: TIL, cheers 😄

inc/packages/namespace.php
rmccue: I'm just a little queasy about relying on that behaviour, but it's fine! Generally I think it's best to handle the default as early as possible.

inc/packages/namespace.php
```diff
@@ -0,0 +1,580 @@
...
+	$types = [ 'synonyms', 'stopwords', 'user-dictionary' ];
+
+	foreach ( $types as $type ) {
+		$text_field = "{$type}-text";
+		$file_field = "{$type}-file";
+		$remove_field = "{$type}-remove";
+		$package_id = null;
+
+		// Handle manual entry.
+		if ( isset( $_POST[ $text_field ] ) && ! empty( $_POST[ $text_field ] ) ) {
```
rmccue, line 286: This is redundant, can be simplified to:

```suggestion
		if ( ! empty( $_POST[ $text_field ] ) ) {
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,580 @@
...
+			// Check for updates.
+			if ( file_exists( $file ) ) {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+				$has_changed = $text !== file_get_contents( $file );
+			}
+
+			// Write to uploads.
+			if ( $has_changed ) {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents
+				file_put_contents( $file, $text );
```
rmccue, line 300: Should check the return code?

inc/packages/namespace.php
```diff
@@ -0,0 +1,580 @@
...
+				if ( is_wp_error( $package_id ) ) {
+					$errors[] = $package_id;
+				}
+			}
+		}
+
+		// Handle file upload.
+		if ( ! empty( $_FILES ) && isset( $_FILES[ $file_field ] ) && ! empty( $_FILES[ $file_field ]['tmp_name'] ) ) {
+			$file = get_package_path( "uploaded-{$type}", $for_network );
+			move_uploaded_file( $_FILES[ $file_field ]['tmp_name'], $file );
```
rmccue, line 311: Should check the return code?

inc/packages/namespace.php
```diff
@@ -0,0 +1,580 @@
...
+		delete_site_option( "altis_search_package_status_{$slug}" );
+		delete_site_option( "altis_search_package_error_{$slug}" );
+	} else {
+		$deleted = delete_option( "altis_search_package_{$slug}" );
+		delete_option( "altis_search_package_status_{$slug}" );
+		delete_option( "altis_search_package_error_{$slug}" );
+	}
+
+	// Update the index settings on successful deletion.
+	if ( $deleted ) {
```
rmccue, line 462: Should be flipped so that the happy path is the main one (and there's less code indented then too, which is #aesthetics)

inc/packages/namespace.php
```diff
@@ -0,0 +1,580 @@
...
+ * @param boolean $update_data Whether the index data should be updated as well.
+ * @return void
+ */
+function do_settings_update( bool $for_network = false, bool $update_data = false ) : void {
+	// Get latest settings.
+	$mapping = Enhanced_Search\elasticpress_mapping( [] );
+	$settings = $mapping['settings'];
+
+	if ( $for_network ) {
+		$sites = Utils\get_sites( 0 );
```
rmccue, line 523: Spicy. What does 0 mean in this context?

rmccue (changes requested):

inc/packages/namespace.php
roborourke: That's just how ElasticPress does it - I think it refers to the network, 0 meaning current network because reasons.

inc/packages/namespace.php
```diff
@@ -0,0 +1,592 @@
...
+				$has_changed = $text !== file_get_contents( $file );
+			}
+
+			// Write to uploads.
+			if ( $has_changed ) {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents
+				$result = file_put_contents( $file, $text );
+				if ( ! $result ) {
+					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					$errors[] = new WP_Error( 'write_package_error', sprintf( __( 'Could not write search package file to %s', 'altis' ), $file ) );
```
rmccue, line 303: Linter error.

inc/packages/namespace.php
```diff
@@ -0,0 +1,592 @@
...
+			}
+		}
+
+		// Handle file upload.
+		if ( ! empty( $_FILES ) && isset( $_FILES[ $file_field ] ) && ! empty( $_FILES[ $file_field ]['tmp_name'] ) ) {
+			$file = get_package_path( "uploaded-{$type}", $for_network );
+			$result = move_uploaded_file( $_FILES[ $file_field ]['tmp_name'], $file );
+			if ( ! $result ) {
+				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+				$errors[] = new WP_Error( 'write_package_error', sprintf( __( 'Could not write search package file to %s', 'altis' ), $file ) );
```
rmccue, line 319: Linter error.

inc/packages/namespace.php
```diff
@@ -0,0 +1,592 @@
...
+				// User dictionary update means we should update the indexed data.
+				if ( $type === 'user-dictionary' ) {
+					$should_update_indexes = true;
+				}
+			}
+		}
+	}
+
+	// Store the errors.
+	array_map( __NAMESPACE__ . '\\add_error_message', $errors );
```
rmccue, line 348: Might be clearer that we're not expecting to use the return value, plus benefits from type hinting at compile time:

```suggestion
	foreach ( $errors as $error ) {
		add_error_message( $error );
	}
```

rmccue (changes requested):

inc/packages/templates/config.php
```diff
@@ -0,0 +1,227 @@
...
+				<?php endif; ?>
+				<p>
+					<input type="file" accept="text/plain" id="stopwords-file" name="stopwords-file" />
+				</p>
+				<h3>
+					<label for="stopwords-text"><?php esc_html_e( 'Manual entry', 'altis' ); ?></label>
+					<?php if ( $types['stopwords']['manual_status'] ) : ?>
+						<span class="es-package-status es-package-status--<?php echo esc_attr( strtolower( $types['stopwords']['manual_status'] ) ); ?>">
+							<?php
+								// translators: %s replaced by status name e.g. 'ACTIVE', 'ERROR'.
```
rmccue, line 207: FWIW, you don't have to say "replaced"; standard format is:
```
// translators: %1$s: status name, %2$s: foo, ...
```

rmccue (approved):

PR #89, opened 2020-07-28
Improve ElasticPress v3 support

roborourke:
This update integrates Altis a little better with ElasticPress v3 in the following ways:

- Ensure ElasticPress won't accidentally delete indexes it doesn't control eg. analytics or .kibana
- Set the posts index name to the same name used in ElasticPress v2 for back compat

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -269,7 +269,7 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
  */
 function on_wp_install() {
 	// This option is used to determine the index name for backwards compat.
-	set_index_version( 1 );
+	set_index_version( 3 );
 }
 
 /**
@@ -285,7 +285,7 @@ function on_delete_index( $query, ?string $type ) {
 	}
 	// Set the version if not already present.
 	if ( ! get_index_version() ) {
-		set_index_version( 1 );
+		set_index_version( 3 );
 	}
 }
 
@@ -328,7 +328,8 @@ function filter_index_name( string $index ) : string {
 		}
 	}
 
-	return $index;
+	// Add ep- prefix to easily determine ElasticPress managed indexes.
+	return "ep-{$index}";
 }
 
 /**
@@ -355,32 +356,10 @@ function filter_documents_pipeline_id( string $id ) : string {
  * @return string
  */
 function protect_non_ep_indexes( string $url, int $failures, string $host, string $path, array $args ) : string {
-
-	// If the path uses a wildcard protect the following index patterns.
-	$protected_index_patterns = [
-		'.kibana',
-		'analytics*',
-	];
-
-	/**
-	 * Filter the protected index patterns. This should be an array of
-	 * Elasticsearch index patterns that ElasticPress is not allowed to
-	 * interact with.
-	 *
-	 * @param array $protected_index_patterns The patterns ElasticPress should ignore.
-	 * @param string $path The remote request path.
-	 * @param array $args The remote request args.
-	 */
-	$protected_index_patterns = apply_filters( 'altis.search.protected_index_patterns', $protected_index_patterns, $path, $args );
-
-	// ElasticPress requests that work on all indexes begin with * so we protect
-	// indexes by name using negation eg. -analytics*,* gives all indexes except
-	// analytics indexes.
-	if ( strpos( $path, '*' ) === 0 ) {
-		$protected_path = array_reduce( $protected_index_patterns, function ( $carry, $pattern ) {
-			return "-{$pattern},{$carry}";
-		}, '' );
-		$path = "{$protected_path}{$path}";
+	// ElasticPress requests that work on all indexes may begin with * so we protect
+	// indexes by enforcing the `ep-` prefix added by our filter.
+	if ( strpos( trim( $path, '/' ), '*' ) === 0 ) {
+		$url = str_replace( '/*', '/ep-*', $url );
 	}
 
 	return $url;
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -85,6 +85,12 @@ function load_elasticpress() {
 	// Modify the default search query to use preset modes.
 	add_filter( 'ep_formatted_args_query', __NAMESPACE__ . '\\enhance_search_query', 10, 2 );
 
+	// Back compat for ElasticPress v2 - change post index name to old version.
+	add_filter( 'ep_index_name', __NAMESPACE__ . '\\filter_index_name' );
+
+	// Ensure non ElasticPress indexes are not affected by global edits using *.
+	add_filter( 'ep_pre_request_url', __NAMESPACE__ . '\\protect_non_ep_indexes', 10, 5 );
+
 	require_once Altis\ROOT_DIR . '/vendor/10up/elasticpress/elasticpress.php';
 
 	// Now ElasticPress has been included, we can remove some of it's filters.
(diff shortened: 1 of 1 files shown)


rmccue:
> Set the posts index name to the same name used in ElasticPress v2 for back compat

Do we want this to be permanent or do we want to migrate to the new naming? Seems like a temporary compat + migration may be nicer for forward compat.

joehoyle:
> Do we want this to be permanent or do we want to migrate to the new naming? Seems like a temporary compat + migration may be nicer for forward compat.

Does ES support just renaming an index?

roborourke:
@rmccue all other indexables will have the slug in the index name, there isn't actually anything to gain from the index name change and the mapping isn't so different that it won't just continue to work between versions.

@joehoyle you can't directly just rename an index, there are a few methods but all essentially involve cloning an existing index and deleting the old one.

I have a plan for using this method to do the zero downtime reindexing too.

rmccue:
> @rmccue all other indexables will have the slug in the index name, there isn't actually anything to gain from the index name change and the mapping isn't so different that it won't just continue to work between versions.

I'm thinking for compatibility with third-party code, consistency with docs, consistency with the other indexes, etc. It seems like long-term and for new sites, it'd be better to use the new-style index name, and treat this as a backwards compatibility thing only.

roborourke:
It can't be done as a back compat thing only unfortunately - it's either change the name of the posts index wholesale via the filter to have back compatibility or require a reindex on upgrade. There's no middle ground while we don't have a migration system that would allow us to run code that adds an alias for example.

Assuming 3rd party code uses the helper functions to get the index name I'm not sure we'd hit any problems there. Regarding documentation we don't have any of our own that mentions the actual index names and I can't see them referenced in the ElasticPress documentation either.

I think it's ok not to have the back compat but realised it was easy to do.

rmccue:
To be clearer what I mean, I'm envisioning something like:

* Add option storing index version, blank is assumed to be v2 indexes. (We'll need to set on install.)
* If the version is marked as v2, filter the index name and change to legacy name. Otherwise, apply no changes to the index name.
* Add migration command to rename the index and/or reindex, and update the option too.

Migration could also be hooked into the WordPress version update if there is one handy we could use from 5.5.

Once we drop v4 support next year, we could remove the now-dead migration code for this.

roborourke:
@rmccue ok, this actually seems to work ok now when switching from v2 to v3, and handles a few other considerations that are the result of upstream bugs. I'll make issues there to fix the problems I've found so far.

There is still an oversight in ElasticPress, when adding a new site the initial content isn't indexed. It's not a blocker per se as this has always been the behaviour but it's not ideal.

Indexing on install is much much nicer now and takes the feature config into account.

inc/namespace.php
```diff
@@ -246,6 +262,130 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
 }
 
+/**
+ * Add default initial options and settings on install.
+ *
+ * @return void
+ */
+function on_wp_install() {
+	// This option is used to determine the index name for backwards compat.
+	set_index_version( 1 );
```
rmccue, line 272: Should this be `3` instead? Thinking it might be good to match the EP version itself.

inc/namespace.php
```diff
@@ -246,6 +262,130 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+	update_option( 'altis_search_index_version', $version );
+}
+
+/**
+ * Get the index version for the current site.
+ *
+ * @return int|null
+ */
+function get_index_version() : ?int {
+	return get_option( 'altis_search_index_version', null );
```
rmccue, line 308: Maybe `?? 1` if we implement suggestion above so this always returns `int`?

inc/namespace.php
```diff
@@ -246,6 +262,130 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+ *
+ * @param string $url The full Elasticsearch request URL.
+ * @param integer $failures Number of failures.
+ * @param string $host Host name.
+ * @param string $path Request path.
+ * @param array $args Remote request arguments.
+ * @return string
+ */
+function protect_non_ep_indexes( string $url, int $failures, string $host, string $path, array $args ) : string {
+
```
rmccue, line 358: Bonus newline:

```suggestion
```

inc/namespace.php
```diff
@@ -246,6 +262,130 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+	/**
+	 * Filter the protected index patterns. This should be an array of
+	 * Elasticsearch index patterns that ElasticPress is not allowed to
+	 * interact with.
+	 *
+	 * @param array $protected_index_patterns The patterns ElasticPress should ignore.
+	 * @param string $path The remote request path.
+	 * @param array $args The remote request args.
+	 */
+	$protected_index_patterns = apply_filters( 'altis.search.protected_index_patterns', $protected_index_patterns, $path, $args );
```
rmccue, line 374: Can we switch this approach to an include instead of an exclude?

inc/namespace.php
roborourke: No, we need something equivalent to the option not being set at all for this to work

inc/namespace.php
roborourke: Hmm, maybe. That would be safer. It would be best if we set a known index prefix for all ElasticPress managed indexes. We could do that using the `EP_PREFIX` constant, or filter I suppose but again it'd require a reindex to work...

I'd want to avoid having to build up a list of all EP indexes by looping over all sites and all indexables to get the index names basically. Will have a think.

inc/namespace.php
roborourke: Hmm, yeah I take your point, though I was thinking of using this value for the zero downtime reindexing.

inc/namespace.php
rmccue: Yeah, I mean `?? 1` here *and* the actual set version as `3` (so `1` is equivalent to unset).

roborourke: Seems a bit weird no? Null seems more meaningful here to me but I think otherwise the fallback value could be 2 if we’re representing the elastic press version

rmccue: 2 seems fine as well yeah. Provided it's installed, the indexes always have a version, it's just whether it's explicit or not, so better to fall back I think.

roborourke: Yep, makes sense. I've updated it now so the default value is 2 for EP v2.

inc/namespace.php
```diff
@@ -197,21 +206,28 @@ function sign_psr7_request( RequestInterface $request ) : RequestInterface {
...
-function log_remote_request_errors( array $request ) {
+function log_remote_request_errors( array $request, ?string $type = null ) {
+	$request_response_body = wp_remote_retrieve_body( $request['request'] );
 	$request_response_code = (int) wp_remote_retrieve_response_code( $request['request'] );
 	$is_valid_res = ( $request_response_code >= 200 && $request_response_code <= 299 );
+	$type = $type ?: 'unknown_request_type';
+
+	// Backup check for errors, sometimes the response is ok but the query
+	// response JSON contains errors.
+	$has_errors = strpos( $request_response_body, '"errors":true' ) !== false;
```
rmccue, line 223: thanks i hate it

inc/namespace.php
```diff
@@ -246,6 +262,109 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+ *
+ * @return void
+ */
+function on_wp_install() {
+	// This option is used to determine the index name for backwards compat.
+	set_index_version( 3 );
+}
+
+/**
+ * When an existing index is deleted set the index version so
```
rmccue, line 276: Forgot to finish this?

inc/namespace.php
```diff
@@ -246,6 +262,109 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+ * @param integer $version The version number.
+ * @return void
+ */
+function set_index_version( int $version ) {
+	update_option( 'altis_search_index_version', $version );
+}
+
+/**
+ * Get the index version for the current site. Defaults to
+ * 2 for ElasticPress version 2.
```
rmccue, line 304: Short desc should be only one sentence ideally.

```suggestion
 * Get the index version for the current site.
 *
 * Defaults to 2 for ElasticPress version 2.
```

inc/namespace.php
```diff
@@ -246,6 +262,109 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+ *
+ * @return int
+ */
+function get_index_version() : int {
+	return get_option( 'altis_search_index_version', null ) ?? 2;
+}
+
+/**
+ * ElasticPress adds the indexable object type to index names. We can maintain backwards
+ * compatibility by filtering the posts indexable index name to remove this type.
```
rmccue, line 314: Add shortdesc to go along with the long one:

```suggestion
 * Replace index name for compatibility.
 *
 * ElasticPress adds the indexable object type to index names. We can maintain backwards
 * compatibility by filtering the posts indexable index name to remove this type.
```

inc/namespace.php
```diff
@@ -246,6 +262,109 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
+function filter_index_name( string $index ) : string {
+	// Back compat for Altis v3 & ElasticPress 2.x
+	// Version 3 of ElasticPress introduces Indexables allowing for user
+	// and term search integration. The new index names follow the pattern
+	// <site>-<indexable>-<blog-id> instead of <site>-<blog-id>.
+	if ( get_index_version() === 2 && strpos( $index, '-post' ) !== false ) {
+		$old_index = str_replace( '-post', '', $index );
+		if ( Elasticsearch::factory()->index_exists( $old_index ) ) {
+			return $old_index;
+		}
```
rmccue, line 328: Should this update the index version if the index doesn't exist?

inc/namespace.php
roborourke: Faster than JSON decoding 😉

inc/namespace.php
roborourke: Hmm maybe yeah. Might actually be better to remove this check, I should've commented at the time why I did it - it was to prevent an error when reindexing from the old name to the new one.

PR #92, opened 2020-08-05
Add ability to register fields for autocompletion

roborourke:
This requires extending the mapping and how documents are analysed at index time so there is an increase in the index size. This however enables the ability to provide autosuggestions and to match partial search phrases. This is useful for autosuggest dropdowns or dynamic searches that update as you type such as in the media library.

### DIFF
--- a/inc/analyzers.json
+++ b/inc/analyzers.json
@@ -116,7 +116,7 @@
 		},
 		"en_stem_filter": {
 			"type": "stemmer",
-			"name": "minimal_english"
+			"name": "light_english"
 		},
 		"es_stop_filter": {
 			"type": "stop",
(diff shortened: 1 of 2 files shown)


inc/analyzers.json
```diff
@@ -116,7 +116,7 @@
 		},
 		"en_stem_filter": {
 			"type": "stemmer",
-			"name": "minimal_english"
+			"name": "light_english"
```
rmccue, line 119: Is this going to require reindexing?

rmccue (approved):

inc/analyzers.json
roborourke: It'll continue to function without reindexing but the results may change a little bit after recreating the indexes. I found this was better.

PR #93, opened 2020-08-24
Update docs to use xargs to process sites individually

nathanielks:
Our internal Cloud docs have recommended this for a minute so I wanted to make sure our user facing docs reflect this!

### DIFF
--- a/docs/reindexing-content.md
+++ b/docs/reindexing-content.md
@@ -17,7 +17,7 @@ wp site list --field=url | xargs -I % wp elasticpress index --url=%
 To re-sychronise the index and update mappings (needed when the mapping has been modified via filters, site language has been changed or a user dictionary has been added and not reindexed successfully):
 
 ```sh
-wp site list --field=url | xargs -I % wp elasticpress index --url=% --setup; wp elasticpress recreate-network-alias
+wp site list --field=url | xargs -I % wp elasticpress index --url=% --setup && wp elasticpress recreate-network-alias
 ```
 
 ## CLI Recommendations
--- a/docs/reindexing-content.md
+++ b/docs/reindexing-content.md
@@ -14,7 +14,7 @@ wp site list --field=url | xargs -I % wp elasticpress index --url=%
 
 `wp site list --field=url` will print each site URL on a new line and ingested by `xargs` to be processed individually.
 
-To re-sychronise the index and update mappings (needed when updating synonym definitions):
+To re-sychronise the index and update mappings (needed when the mapping has been modified via filters, site language has been changed or a user dictionary has been added and not reindexed successfully):
 
 ```sh
 wp site list --field=url | xargs -I % wp elasticpress index --url=% --setup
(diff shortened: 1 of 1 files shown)


roborourke:
The `--network-wide` flag has other effects we should take into account such as creating the network wide search alias. I'd like to double check that this won't miss anything out that the plugin could be expecting to be present.

nathanielks:
@roborourke is that something you have the bandwidth for or do need help with that?

nathanielks:
Rather, if it's at all possible for you to investigate that further, it would be a significant help to me, otherwise I'm happy to do the research!

docs/reindexing-content.md
```diff
@@ -6,24 +6,26 @@ When making changes the data or analyzers that are stored in the search index, y
...
 
 ```sh
-wp elasticpress index --network-wide
+wp site list --field=url | xargs -I % wp elasticpress index --url=%
 ```
 
-Re-sychronise the index and updating mappings (needed when updating synonym definitions)
+`wp site list --field=url` will print each site URL on a new line and ingested by `xargs` to be processed individually.
+
+To re-sychronise the index and update mappings (needed when updating synonym definitions):
```
roborourke, line 17:
```suggestion
To re-sychronise the index and update mappings (needed when the mapping has been modified via filters, site language has been changed or a user dictionary has been added and not reindexed successfully):
```

docs/reindexing-content.md
```diff
@@ -6,24 +6,26 @@ When making changes the data or analyzers that are stored in the search index, y
...
 
 ```sh
-wp elasticpress index --setup
+wp site list --field=url | xargs -I % wp elasticpress index --url=% --setup
```
roborourke, line 20:
```suggestion
wp site list --field=url | xargs -I % wp elasticpress index --url=% --setup && wp elasticpress recreate-network-alias
```

roborourke (changes requested):
When doing the set up without the `--network-wide` flag the main thing missed is the network alias creation, there is a CLi command that can be tacked on to handle this though. Added as a suggestion but not sure if `xargs` will handle the `&&` as added here.

docs/reindexing-content.md
nathanielks: Does this need to be run once or for each network? If only once, your recommendation is sound! Bash picks up the `&&` before `xargs` does, so it's treated as a separate command from `xargs`.

roborourke: Ah that's what I was worried about, needs to be scoped to the xargs command

nathanielks: Looking at the source, it appears to loop through all sites, so I believe this should run outside of the xargs command (how this PR has it currently written). What do you think?

2.8.2: https://github.com/10up/ElasticPress/blob/b56e8f8dc2f07cd6c591edc5db94eb49236bc549/bin/wp-cli.php#L282-L310

latest: https://github.com/10up/ElasticPress/blob/c8c686d49f588d8ccfdc4e1cf0b22d9459f8276e/includes/classes/Command.php#L456-L486

roborourke: Missed that bit! Alrighty then, we're good here

roborourke (approved):

PR #96, opened 2020-09-03
Add fuzzy search config

### LINKED ISSUE #95: Add configuration option for search fuzziness level
The search defaults to using automatic fuzziness but this should be made configurable. Given that when using the advanced search mode fuzziness is applied to each term rather than the whole phrase it can still generate a lot of results that aren't completely useful.

roborourke:
Add support for setting edit distance, prefix length, max expansions and transpositions values

Fixes #95

### DIFF
--- a/docs/search-configuration.md
+++ b/docs/search-configuration.md
@@ -79,3 +79,62 @@ By default the post title, excerpt, content, author name and taxonomy terms are 
 ```
 
 It is important to note that the field names should match the fields in the elasticsearch index and do not always correspond 1:1 with database fields.
+
+## Fuzzy Matching
+By default some degree of fuzzy matching is allowed so that simple spelling errors can still return some results, for example "breif" would match "brief". This can sometimes result in unwanted search results however with short words and acronyms.
+
+Fuzzy matching works by providing an edit distance as an integer from 0-2. The number indicates how many edits are allowed for a term to match. Edits can be one of the following:
+
+* Changing a character (box → fox)
+* Removing a character (click → lick)
+* Inserting a character (sic → sick)
+* Transposing two adjacent characters (act → cat)
+
+**Note:** when using "advanced" search mode quoted strings will not use fuzzy matching.
+
+**Note:** if you have synonyms configured they will not use fuzzy matching.
+
+Below is the default fuzzy search configuration.
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"search": {
+					"fuzziness": {
+						"distance": "auto:4,7",
+						"prefix-length": 1,
+						"max-expansions": 40,
+						"transpositions": true
+					}
+				}
+			}
+		}
+	}
+}
+```
+
+**`distance`**
+
+This determines the edit distance for fuzzy matching. This value can also be set as the value for `fuzziness` directly if you do not need to edit the other options.
+
+- If an integer is provided this is used as a fixed edit distance eg. 0, 1 or 2 edits allowed
+- If a string is provided it must be in the form `auto:[min],[max]`
+  - If `min` is 3 all terms from 0-2 characters long will have an edit distance of 0
+  - If `max` is 6 all terms from `min` to 6 characters long will have an edit distance of 1
+  - Any terms longer than `max` will have an edit distance of 2
+
+**`prefix-length`**
+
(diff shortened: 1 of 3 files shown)


joehoyle:
This looks pretty good. In a lot of these changes, I keep wondering how we are different to default search. In mysql I'm assuming these is no fuzzy searching at all, but does it have partial word matching, or does it try to stick to word boundaries?

inc/namespace.php
```diff
@@ -941,12 +987,18 @@ function enhance_search_query( array $query, array $args, string $type = 'post'
...
+		// - 7+ chars = 2 edits.
+		$query['bool']['should'][1]['multi_match']['fuzziness'] = $fuzziness['distance'];
+		$query['bool']['should'][1]['multi_match']['prefix_length'] = $fuzziness['prefix-length'];
+		$query['bool']['should'][1]['multi_match']['max_expansions'] = $fuzziness['max-expansions'];
+		$query['bool']['should'][1]['multi_match']['fuzzy_transpositions'] = $fuzziness['transpositions'];
 	}
 
 	if ( $mode === 'advanced' ) {
 		$query['bool']['should'] = [
-			get_advanced_query( $args['s'], $search_fields, $strict ),
```
joehoyle, line 949: Was / is this a bug in v4?

docs/search-configuration.md
```diff
@@ -79,3 +79,62 @@ By default the post title, excerpt, content, author name and taxonomy terms are
...
+	}
+}
+```
+
+**`distance`**
+
+This determines the edit distance for fuzzy matching. This value can also be set as the value for `fuzziness` directly if you do not need to edit the other options.
+
+- If an integer is provided this is used as a fixed edit distance eg. 0, 1 or 2 edits allowed
+- If a string is provided it must be in the form `auto:[min],[max]`
```
rmccue, line 123: Are we using the `auto:` prefix because it's part of ES or something else? Is it just future-proofing?

inc/namespace.php
```diff
@@ -892,6 +892,51 @@ function filter_facet_settings( $value ) {
...
+	$fuzziness = wp_parse_args( $fuzziness, [
+		'distance' => 'auto:4,7',
+		'prefix-length' => 1,
+		'max-expansions' => 50,
+		'transpositions' => true,
+	] );
+
+	// Validate distance value.
+	if ( ! preg_match( '/^([0-2]|auto:\d+,\d+|auto)$/', $fuzziness['distance'] ) ) {
+		trigger_error( sprintf( 'The provided fuzziness distance config option %s is invalid, defaulting to "auto:4,7"', (string) $fuzziness ), E_USER_WARNING );
```
rmccue, line 923: This doesn't need to be cast to string, as the `%s` formatter will already do this. That said, need to ignore the linter error here too.

rmccue (changes requested):

inc/namespace.php
roborourke: No, it's ultimately to pass more useful data to the new filter in the `get_advanced_query()` function. I was looking at using the existing `ep_fuzziness_arg` filter in that function (which needs the whole `$args` array) but opted for providing more control than that offers.

docs/search-configuration.md
roborourke: Part of ES https://www.elastic.co/guide/en/elasticsearch/reference/6.8/common-options.html#fuzziness

roborourke:
@joehoyle correct MySQL has no fuzzy matching. It creates a relevancy score based on the fields matched eg. title, excerpt & content and whether it was the whole phrase or a partial match. It does tokenise the search term into words and uses them in a `LIKE` query eg. `%<word>%`.

I'd say the key differentiators to default search are:

- Searching against author, terms, categories and other extra fields vs title, excerpt and content only
- Date based relevancy decay
- Word stemming
- Stopwords
- Fuzzy matching
- Synonyms
- User dictionary (for Japanese)

We could document those in a "Why this is good" section. I think there's a case for highlighting the business and technical value of ES as a solution for search.

docs/search-configuration.md
roborourke: I toyed with some other formats but it felt like an unnecessary abstraction at that point.

rmccue: 👍 can we link the ES docs directly from here so it's clearer that you're setting ES config? Will help with being able to explore how to tune the settings I think.

rmccue (approved):

PR #98, opened 2020-09-10
Fix mime type search queries

### LINKED ISSUE #97: Media Manager returning empty results when searching
The WP Media Manager returns an empty array when a search is done.

Steps to reproduce (Elastisearch needs to be enabled):
1. Open an existing/create a new post.
2. Insert an image block
3. Click on Media Library
4. Search for anything
5. Results are empty.

This doesn't happen when the same search is done from the Media Library screen or when the search is empty:

- If the search is empty, ElasticPress will disable itself for that search
- A search done by the Media Manager from Gutenberg sends an extra parameter (mime type) and here's where the problem seems to be.

Here's the query I'm triggering in ES (change `"query": "beach" for something that could work for you):

```json
{
  "from": 0,
  "size": 40,
  "sort": [
    {
      "post_date": {
        "order": "desc"
      }
    }
  ],
  "query": {
    "function_score": {
      "query": {
        "bool": {
          "should": [
            {
              "simple_query_string": {
                "query": "beach",
                "fields": [
                  "post_title",
                  "post_excerpt",
                  "post_content",
                  "alt",
                  "rekognition_labels",
                  "attachments.attachment.content"
                ],
                "default_operator": "and"
              }
            }
          ]
        }
      },
      "exp": {
        "post_date_gmt": {
          "scale": "14d",
          "decay": 0.25,
          "offset": "7d"
        }
      },
      "score_mode": "avg",
      "boost_mode": "sum"
    }
  },
  "post_filter": {
    "bool": {
      "must": [
        {
          "terms": {
            "post_mime_type": [
              "image"
            ]
          }
        },
        {
          "term": {
            "post_type.raw": "attachment"
          }
        },
        {
          "terms": {
            "post_status": [
              "inherit",
              "private"
            ]
          }
        }
      ]
    }
  },
  "explain": true
}
```

Note that removing
```json
{
          "terms": {
            "post_mime_type": [
              "image"
            ]
          }
        },
```

would return results with no problem.

Acceptance criteria:

- [ ] Should we filter by mime type? I don't think there's a big reason to filter by type when we're searching. This would only make sense if the user has an available selector to filter by mime type but that's not the case.

roborourke:
ElasticPress has a bug that results in searches incorrectly filtering on exact matches for truncated mime types such as `image` meaning no results are returned. This corrects the use of terms and also adds prefix queries which are more performant than the built regex query.

Fixes #97

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -88,6 +88,9 @@ function load_elasticpress() {
 	add_filter( 'ep_term_formatted_args_query', __NAMESPACE__ . '\\enhance_term_search_query', 10, 2 );
 	add_filter( 'ep_user_formatted_args_query', __NAMESPACE__ . '\\enhance_user_search_query', 10, 2 );
 
+	// Fix the mime type search query.
+	add_filter( 'ep_formatted_args', __NAMESPACE__ . '\\fix_mime_type_query', 11, 2 );
+
 	// Back compat for ElasticPress v2 - change post index name to old version.
 	add_filter( 'ep_index_name', __NAMESPACE__ . '\\filter_index_name' );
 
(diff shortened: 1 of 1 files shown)


roborourke:
I'll make an upstream issue too

roborourke:
Upstream issue here https://github.com/10up/ElasticPress/issues/1604

inc/namespace.php
```diff
@@ -268,6 +271,86 @@ function noop_wp_query_found_rows_on_failed_ep_request( string $sql, WP_Query $q
...
 
+/**
+ * Modify the mime type query to support truncated forms.
+ *
+ * @param array $query The Elasticsearch query.
+ * @param array $args The WP_Query query vars.
+ * @return array
+ */
+function fix_mime_type_query( array $query, array $args ) : array {
+	if ( empty( $args['post_mime_type'] ) ) {
```
igmoweb, line 282: Just a simplification: Join both ifs:
```suggestion
	if ( empty( $args['post_mime_type'] ) || ! isset( $query['post_filter'] ) ) {
```

igmoweb (approved):

inc/namespace.php
roborourke: More terse but I find things like this tend to be personal preference. Cheers the review though, will get this backported asap

PR #105, opened 2020-09-23
Add updated docs for Altis v5

roborourke:
This includes more details on how the CMS queries are integrated, how to use autosuggest search and the user dictionaries feature.


docs/README.md
```diff
@@ -2,7 +2,7 @@
 
 ![](./assets/banner-search.png)
 
-The Search module provides a mirrored index of all CMS content that is optimized for search relevance, speed and accuracy. The Search module overrides the default search functionality to query the specialized search index in place of a standard MySQL query. This means all default search operations using the CMS search APIs for posts such as `WP_Query`, `get_posts`, `get_search_form()` and the REST API search endpoints will transparently make use of the search index.
+The Search module provides a mirrored Elasticsearch index of all CMS content that is optimized for search relevance, speed and accuracy.
```
rmccue, line 5: Do we want to mention something about it differing from MySQL?

```suggestion
The Search module provides a mirrored Elasticsearch index of all CMS content that is optimized for search relevance, speed and accuracy. This operates at a higher level than the primary datastore (MySQL) and is the recommended solution for searching.
```

or something like that; we mention queries can be done in MySQL below, so seems we should have something here

docs/cms-query-integration.md
```diff
@@ -0,0 +1,64 @@
...
+## Autosuggest
+Because of the way Elasticsearch works it performs better with complete search terms as you might expect to be submitted from a search form.
+
+In some cases you might want to provide a dynamic interface where results are fetched as the user types. The standard method of search will not produce good results until a full word has been typed out so Altis provides a means of analysing and fetching results based on partial search terms.
+
+In the same way `ep_integrate` can be passed to any query class you can pass the parameter `autosuggest`:
+
+```php
+$posts = new WP_Query( [
+	'autosuggest' => true,
```
rmccue, line 47: Is `ep_integrate` required too? We should state explicitly either way.

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,74 @@
+# Custom User Dictionaries
+
+The search module supports 3 types of custom dictionaries for helping to tune search results to your content. These are synonyms, stop words and custom text analysis for Japanese.
+
+In the website admin you can upload text files (`.txt` only) or in the case of stop words and synonyms manually enter them on the [Search Configuration](admin://admin.php?page=search-config) screen.
+
+Note that only a file upload is supported for Japanese user dictionaries.
```
rmccue, line 7: Interesting; why?

rmccue (changes requested):

docs/README.md
```diff
@@ -78,7 +92,5 @@ Facets are a feature in ElasticPress which add control to filter content by one
 
 Depending on the configuration specified for `facets`, if the `match-type` property is set to `any`, it will force the results to match any selected taxonomy term. If set to `all`, it will match to results with all of the selected terms.
 
-### Auto Suggest
-The default auto suggest functionality has been modified but does not effect the default WP search form template.
-
-In addition to the default auto suggest endpoint, an additional endpoint was added (`/autosuggest/`). This endpoint accepts as json object to modify the parameters that are forwarded to ElasticPress for suggesting posts.
+### Search Form Auto Suggest
+This feature enhances search forms on the website to show a dropdown list of suggestions as users type.
```
rmccue, line 96: Does this also need a link to some relevant documentation?

roborourke: I realised that when this was implemented the code I wrote to make this secure for Enterprise Kit wasn't actually copied over. I'm somewhat tempted to remove or replace the built in ElasticPress feature altogether to be honest. We don't have any other documentation for it right now anyway and we didn't have anything more detailed previously.

docs/search-configuration/custom-dictionaries.md
roborourke: The file upload and the manual entry count as separate files, this is easy to support for stopwords and synonyms but for the user dictionary there can only be a single tokeniser and a single user dictionary file in the settings for it. I can add that information to the docs here.

docs/cms-query-integration.md
roborourke: Ah yeah, I'll make it clearer that `ep_integrate` defaults to true when the `s` parameter is used.

docs/README.md
roborourke: Yeah. I'll add more info as to the benefits of ES as well.

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
+# CMS Query Integration
+
+The Search module overrides the default search functionality to query the specialized search index in place of a standard MySQL query. This means all default search operations using the CMS search APIs for posts, users and terms such as `WP_Query`, `WP_User_Query`, `WP_Term_Query` and the REST API search endpoints will transparently make use of the search index by default.
+
+It is possible to control this behaviour and thus use the search index for non search queries that might otherwise be too slow to do in the database such as meta queries.
```
jazzsequence, line 5: "non search" should be hyphenated, and there should be commas.

```suggestion
It is possible to control this behaviour, and thus use the search index for non-search queries that might otherwise be too slow to do in the database, such as meta queries.
```

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+
+Each query class accepts an array of parameters. ElasticPress checks for the value of an `ep_integrate` parameter, if this is true the query is performed by Elasticsearch instead of MySQL. If the `s` (search) parameter is present `ep_integrate` will be set to `true` automatically.
```
jazzsequence, line 7: Second sentence can be broken up more.

```suggestion
Each query class accepts an array of parameters. ElasticPress checks for the value of an `ep_integrate` parameter. If this is true, the query is performed by Elasticsearch instead of MySQL. If the `s` (search) parameter is present `ep_integrate` will be set to `true` automatically.
```

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+
+The following example sets `ep_integrate` to true for a non search query, that is any query where the `s` parameter is not used.
```
jazzsequence, line 9: Hyphenate "non search", add period, add comma after "that is"

```suggestion
The following example sets `ep_integrate` to true for a non-search query. That is, any query where the `s` parameter is not used.
```

I'm not sure i like the sentence starting with "That is", so perhaps this sentence can be shortened to:

```
The following example sets `ep_integrate` to true for a non-search query where the `s` parameter is not used.
```

or

```
The following example sets `ep_integrate` to true for a query where the `s` parameter is not used.
```

Since this is a "for example", it should probably end with a `:` instead of a period.

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+		[
+			'key' => 'color',
+			'value' => [ 'blue', 'green' ],
+			'compare' => 'IN'
+		],
+	],
+] );
+```
+
+Note in the above example you would need a PHP Code Sniffer exception for using the meta query to pass [Automated Code Review](docs://guides/code-review/README.md).
```
jazzsequence, line 25: Comma after "in the above example." Also, if a PHPCS ignore is required, why not add it to the example (as has been done in other docs)? And then explain why the ignore is there instead.

```suggestion
Note in the above example, you would need a PHP Code Sniffer exception for using the meta query to pass [Automated Code Review](docs://guides/code-review/README.md).
```

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+	if ( ! $query->is_search() || $query->get( 'post_type' ) !== 'project' ) {
+		return;
+	}
+
+	$query->set( 'ep_integrate', false );
+}, 20 );
+```
+
+## Autosuggest
+Because of the way Elasticsearch works it performs better with complete search terms as you might expect to be submitted from a search form.
```
jazzsequence, line 41: Comma

However, this opens up questions about "the way Elasticsearch works", e.g. "_how_ does it work?" This sentence assumes pre-existing knowledge or information about "the way Elasticsearch works". Is there relevant documentation that can be linked to that describes this?

```suggestion
Because of the way Elasticsearch works, it performs better with complete search terms as you might expect to be submitted from a search form.
```

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+
+In some cases you might want to provide a dynamic interface where results are fetched as the user types. The standard method of search will not produce good results until a full word has been typed out so Altis provides a means of analysing and fetching results based on partial search terms.
```
jazzsequence, line 43: Comma

```suggestion
In some cases, you might want to provide a dynamic interface where results are fetched as the user types. The standard method of search will not produce good results until a full word has been typed out, so Altis provides a means of analysing and fetching results based on partial search terms.
```

Is this the explanation to the above "way that ES works"?

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+
+In the same way as `ep_integrate` can be passed to any query class you can pass the parameter `autosuggest` along with the `s` parameter:
```
jazzsequence, line 45: Comma

```suggestion
In the same way as `ep_integrate` can be passed to any query class, you can pass the parameter `autosuggest` along with the `s` parameter:
```

docs/cms-query-integration.md
```diff
@@ -0,0 +1,68 @@
...
+	// Setting `s` will cause `ep_integrate` to be added and set to true.
+	's' => 'str',
+] );
+```
+
+The `autosuggest` parameter defaults to true if the `DOING_AJAX` constant is defined and `true`.
+
+### Filters
+
+By default only a subset of fields are analysed for autosuggest searches but this can be filtered to add additional fields for each indexable object type.
```
jazzsequence, line 60: Comma

```suggestion
By default only a subset of fields are analysed for autosuggest searches, but this can be filtered to add additional fields for each indexable object type.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
+# Custom User Dictionaries
+
+The search module supports 3 types of custom dictionaries for helping to tune search results for your content. These are synonyms, stop words and custom text analysis for Japanese.
+
+In the website admin you can upload text files (`.txt` only) and in the case of stop words and synonyms also manually enter them on the [Search Configuration](admin://admin.php?page=search-config) screen.
```
jazzsequence, line 5: Commas

```suggestion
In the website admin, you can upload text files (`.txt` only) and, in the case of stop words and synonyms, also manually enter them on the [Search Configuration](admin://admin.php?page=search-config) screen.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+Note that only a file upload is supported for the Japanese user dictionary as it is used by the [kuromoji tokenizer](https://www.elastic.co/guide/en/elasticsearch/plugins/6.3/analysis-kuromoji-tokenizer.html) that splits words up during analysis. Synonyms and stop words are "filters" so multiple files (uploaded and manually entered) can be used for those.
```
jazzsequence, line 7: Comma

```suggestion
Note that only a file upload is supported for the Japanese user dictionary, as it is used by the [kuromoji tokenizer](https://www.elastic.co/guide/en/elasticsearch/plugins/6.3/analysis-kuromoji-tokenizer.html) that splits words up during analysis. Synonyms and stop words are "filters" so multiple files (uploaded and manually entered) can be used for those.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+User dictionaries can be configured for [the entire network](admin://network/admin.php?page=search-config) (provided the subsites match the primary site language) or at an individual site level.
+
+## Synonyms
+For better search relevance and results, a dictionary of synonyms can be very useful in the following cases:
+
+- There are a lot of acronyms or other contractions in your content
+- There are common mis-spellings of search terms
+- There are users from different countries who may use different words for the same thing
+- There are sub categories of content that should match a more general search term
```
jazzsequence, line 17: "Sub categories" should be one word or hyphenated

```suggestion
- There are sub-categories of content that should match a more general search term
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+Note that only a file upload is supported for the Japanese user dictionary as it is used by the [kuromoji tokenizer](https://www.elastic.co/guide/en/elasticsearch/plugins/6.3/analysis-kuromoji-tokenizer.html) that splits words up during analysis. Synonyms and stop words are "filters" so multiple files (uploaded and manually entered) can be used for those.
+
+User dictionaries can be configured for [the entire network](admin://network/admin.php?page=search-config) (provided the subsites match the primary site language) or at an individual site level.
+
+## Synonyms
+For better search relevance and results, a dictionary of synonyms can be very useful in the following cases:
+
+- There are a lot of acronyms or other contractions in your content
+- There are common mis-spellings of search terms
```
jazzsequence, line 15: "mis-spellings" doesn't need to be hyphenated (I don't think? Is that a UK English/US English thing?)

```suggestion
- There are common misspellings of search terms
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+- There are users from different countries who may use different words for the same thing
+- There are sub categories of content that should match a more general search term
+
+For example, you may want search results for the phrase "Checking Accounts" to match content that is titled "Current Accounts" or for common acronyms like "GDP" to match "Gross Domestic Product".
+
+Additionally to cater for variations in language you could create synonyms for American English e.g. "sneakers" to British English e.g. "trainers".
```
jazzsequence, line 21: Commas

```suggestion
Additionally, to cater for variations in language, you could create synonyms for American English e.g. "sneakers" to British English e.g. "trainers".
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+Comma separated lists of words are treated as equivalent. You should have one list of synonyms per line.
```
jazzsequence, line 23: "Comma separated" should be hyphenated

```suggestion
Comma-separated lists of words are treated as equivalent. You should have one list of synonyms per line.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+sneakers, trainers, footwear, shoes
+foozball, foosball, table football
+CPU, central processing unit
+```
+
+Comma separated words or phrases followed by "=>" will be treated the same as comma separated words or phrases to the right of the "=>" operator but not the other way around.
```
jazzsequence, line 31: Hyphen here too

```suggestion
Comma-separated words or phrases followed by "=>" will be treated the same as comma-separated words or phrases to the right of the "=>" operator but not the other way around.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+i-pod, i pod => ipod
+tent => bivouac, teepee
+sea biscuit, sea biscit => seabiscuit
+```
+
+In the above example a search for "tent" will match content containing "bivouac" or "teepee" but a search for "teepee" will only return results containing the word "teepee" specifically.
```
jazzsequence, line 39: Comma

```suggestion
In the above example, a search for "tent" will match content containing "bivouac" or "teepee" but a search for "teepee" will only return results containing the word "teepee" specifically.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+ignore
+these
+words
+```
+
+
+## Japanese User Dictionary
+Your site's language must be set to Japanese to see this option.
+
+A user dictionary provides a way to control how words are broken up when searching. If there are compound words or phrases specific to this site such as the names of authors or celebrities that users may search for they can be specified here to increase search relevancy.
```
jazzsequence, line 59: Commas here

```suggestion
A user dictionary provides a way to control how words are broken up when searching. If there are compound words or phrases specific to this site, such as the names of authors or celebrities that users may search for, they can be specified here to increase search relevancy.
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -0,0 +1,73 @@
...
+
+The syntax for the provided text file should follow the CSV format:
+
+```csv
+text, token 1 ... token n, reading 1 ... reading n, part-of-speech tag
```
jazzsequence, line 64: This example is obscure and not really explained (without knowing Japanese). Is there a way to better describe the format/syntax of the CSV file, specifically what `text`, `token`, `reading` and `part-of-speech tag` refer to?

jazzsequence (changes requested):
Some punctuation suggestions but also a few more general questions about the documentation itself

docs/search-configuration/custom-dictionaries.md
roborourke: I don't really understand it myself to be honest. Just hoping that between this and the actual Japanese example below it'll make sense to those who need to use it.

jazzsequence: I wonder if we could pull someone from APAC who has a better handle on it to suggest changes? I hate to leave something like that and then, in the future if we're working with partner devs or people not familiar with the Japanese dictionaries look at our docs and go "huh?" and we're like ¯\_(ツ)_/¯

docs/README.md
```diff
@@ -22,18 +34,37 @@ If you do not wish to use the search module it can be deactivated via your confi
...
 
 When used in conjunction with the [Media Rekognition](docs://media/image-recognition.md) feature, all images are processed for automatic keyword detection and stored in the search index too.
 
+## CMS Query Integration
+The method for determining whether CMS queries are handled by Elasticsearch or MySQL uses sensible defaults but can also be controlled in more granular ways through code. In addition there is special handling required for supporting partial search terms common in autosuggest or "typeahead" interfaces.
+
+[Find out how to control Elasticsearch query integration and how to use autosuggest search here](./cms-query-integration.md).
+
+## Search Configuration
+The default search behaviour can be tuned to allow for stricter or more permissive matching as well as enabling advanced search query capabilities such as using quoted strings for exact matches. You can also tune the relevancy of specific fields, and how fuzzy matching works.
```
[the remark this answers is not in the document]
roborourke: Good catch - US english all the way with public facing stuff.

docs/README.md
```diff
@@ -2,7 +2,19 @@
...
+- Advanced text analysis
+  - Removing HTML
+  - Language dependent word stemming and stopwords
+  - Language dependent tokenisation
+- Configurable relevancy scores
+- Fuzzy matching
+- Synonyms
+- Aggregations for faceted search and statistics
 
 The default Search index and related functionality is provided by the [ElasticPress plugin](https://github.com/10up/ElasticPress) and the multilingual support is derived from [the WordPress.com ElasticSearch library](https://github.com/Automattic/wpes-lib).
```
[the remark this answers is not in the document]
roborourke:
```suggestion
The default Search index and related functionality is provided by the [ElasticPress plugin](https://github.com/10up/ElasticPress) and the multilingual support is derived from [the WordPress.com Elasticsearch library](https://github.com/Automattic/wpes-lib).
```

docs/README.md
```diff
@@ -2,7 +2,19 @@
...
```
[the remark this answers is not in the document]
roborourke: Not sure what you mean here about hierarchy?

docs/README.md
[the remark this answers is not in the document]
jazzsequence: That's what I thought. There's another "behaviour" somewhere else, too, IIRC

docs/README.md
```diff
@@ -22,18 +34,37 @@ If you do not wish to use the search module it can be deactivated via your confi
...
+- Post Author
+- Users
+- User Meta
+- Terms
+- Term Meta
 
 When used in conjunction with the [Media Rekognition](docs://media/image-recognition.md) feature, all images are processed for automatic keyword detection and stored in the search index too.
 
+## CMS Query Integration
+The method for determining whether CMS queries are handled by Elasticsearch or MySQL uses sensible defaults but can also be controlled in more granular ways through code. In addition there is special handling required for supporting partial search terms common in autosuggest or "typeahead" interfaces.
```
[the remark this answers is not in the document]
roborourke:
```suggestion
The method for determining whether CMS queries are handled by Elasticsearch or MySQL uses sensible defaults, but can also be controlled in more granular ways through code. Special handling is required for supporting partial search terms common in autosuggest or "typeahead" interfaces.
```

docs/cms-query-integration.md
roborourke: The second comma after database reads a bit funny to me, suggesting this instead:
```suggestion
It is possible to control this behaviour and thus use the search index for non-search queries. The reason for doing this is to improve performance when running complex queries, such as meta queries, that would otherwise be too slow to do in the database.
```

docs/cms-query-integration.md
roborourke: I like the first replacement sentence example, good shout. I think it's important to call out "non-search" still

docs/cms-query-integration.md
jazzsequence: LGTM

docs/README.md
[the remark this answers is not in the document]
jazzsequence: This was referring to the example config block which is actually below this line, but I couldn't suggest changes on it because it didn't actually change.

docs/cms-query-integration.md
roborourke: 🤔 yeah. To some extent we do expect a bit of knowledge but perhaps we can avoid that wording altogether, eg.
```suggestion
Because of the default way that Elasticsearch analyses text it performs better with complete search terms, as you might expect to be submitted from a search form.
```

docs/cms-query-integration.md
roborourke: Not quite, the "way that ES works" was referring to how text is analysed, so how sentences are broken up, how words are stemmed or ignored and so on. To enable typeahead functionality we have to add an additional method of text analysis called [ngrams](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenfilter.html) that split words up into chunks rather than doing anything more clever.

docs/cms-query-integration.md
roborourke:
```suggestion
The following example sets `ep_integrate` to true for a non-search query where the `s` parameter is not used.
```

docs/cms-query-integration.md
roborourke: Not sure this one is necessary myself but have noticed that I've been a bit lazy with that sentence.
```suggestion
Note that in the above example you would need a PHP Code Sniffer exception for using the meta query to pass [Automated Code Review](docs://guides/code-review/README.md).
```

roborourke: Good point re. adding the phpcs rule too. I'll let into what that is.

docs/cms-query-integration.md
roborourke: Don't think a comma after "cases" is needed. At least when I read it in my head I don't pause naturally there. Suggesting:
```suggestion
In some cases you might want to provide a dynamic interface where results are fetched as the user types. The standard method of search will not produce good results until a full word has been typed out, so Altis provides a means of analysing and fetching results based on partial search terms.
```

docs/cms-query-integration.md
jazzsequence: I like that phrasing better, but I'd remove `default`. Don't forget the comma (in the right place) ;p

```suggestion
Because of the way Elasticsearch analyzes text, it performs better with complete search terms as you might expect to be submitted from a search form.
```

docs/search-configuration/custom-dictionaries.md
roborourke: No idea tbh, ironically I'm not fussy on how it's spelled 😄

docs/cms-query-integration.md
jazzsequence: ...says the person who doesn't put a comma after "at least" 😂

I think it's probably fine to leave out the comma there. It works either way, for the most part.

docs/README.md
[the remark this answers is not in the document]
roborourke: Ah gotcha - I think you can expand the preview and select multiple lines. Lets spin up a platform-dev issue. Maybe we need to write an RFC for documentation standards that goes into real detail around config and code examples etc.

docs/search-configuration/custom-dictionaries.md
roborourke: @Kenshino do you understand what the above line of documentation is referring to? Do you think most Japanese speakers would?

docs/README.md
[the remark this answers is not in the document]
jazzsequence: Yep, it's definitely a broader conversation 👍

docs/search-configuration/custom-dictionaries.md
Kenshino: I don't understand it myself. I don't think Japanese people would @roborourke

jazzsequence: For reference, here's the full proposed documentation block. I'm not sure it helps that much, but just for context:

## Japanese User Dictionary
Your site's language must be set to Japanese to see this option.

A user dictionary provides a way to control how words are broken up when searching. If there are compound words or phrases specific to this site, such as the names of authors or celebrities that users may search for, they can be specified here to increase search relevancy.

The syntax for the provided text file should follow the CSV format:

```csv
text, token 1 ... token n, reading 1 ... reading n, part-of-speech tag
```

For example:

```csv
東京スカイツリー,東京 スカイツリー,トウキョウ スカイツリー,カスタム名詞
```

Will cause "東京スカイツリー" to be interpreted as "東京" and "スカイツリー" rather than "東京", "スカイ" and "ツリ".

roborourke: If I'm following correctly in the above example:

1. `text` is a string of text that can appear in the website content
2. `token 1...n` is the same text as 1 but with spaces added where the text should be split up
3. `reading 1...n` is the same text as 2 (keeping the spaces) but with kanji written as katakana instead
4. `part-of-speech` indicates whether the text is a noun, adjective, verb and so on

The valid part of speech tags are documented here https://www.kanshudo.com/grammar/parts_of_speech but I can list them in the docs here too. Some examples:

- カスタム名詞 / custom noun
- めいし / 名詞 / noun
- こゆうめし / 固有名詞 / proper noun
- けいしきめいし / 形式名詞 / nominalizer
   > A nominalizer is a word added to a verb or adjective that turns it into a noun phrase, a very common pattern in Japanese, which contains several nominalizers including の, もの, こと etc. See 形式名詞（けいしきめいし).
- どうし /  動詞 / verb
- じょどうし / 助動詞 / auxiliary verb
- ふくごうどうし / 複合動詞 / compound verb
- けいようし / 形容詞 / adjective; い adjective
- けいようどうし / 形容動詞 / な / adjective
  > The origin of the term 形容動詞, which literally means 'verb adjective', is interesting. Originally, な adjectives emerged through the use of なる (become) to a noun (1) - for example 
  > いだい / 偉大
  > なる
  > かわ
  > 川
  > So they were literally 'adjectives formed through addition of a verb'. Over time, 'なる' became just 'な’.
- れんたいし / 連体詞 / other (not い or な) adjective
- ぜんちし / 前置詞 / preposition
- じしょけい / 辞書形 / dictionary form
   > The form that appears in a dictionary. For a verb, this is the plain form of the present tense. For a noun or adjective, it is the uninflected form.
- ふつうけい / 普通形 / plain form(s) of a word
   > The plain forms are the 'casual' or 'short' forms. For example, for the verb 
   > た
   > 食
   > べる (eat), 食べる, 食べた, 食べている etc. are all plain forms.
- ていねいけい / 丁寧形 / polite form(s) of a word
- かつようけい / 活用形 / conjugated form of a word
- しゅうしけい / 終止形 / predicative form
- めいれいけい / 命令形 / imperative form
- しえきけい / 使役形 / causative form
- よだんかつよう / 四段活用 / classical Japanese verb form
- じょうけんけい / 条件形 / conditional form
- じょすうし / 助数詞 / counter
- たいひ / 対比 / comparison
- ひてい / 否定 / negative
- きょうちょう / 強調 / emphasis
- かんりゃくか / 簡略化 / simplification (of a kanji etc.)
- ひっしゃ / 筆者 / cursive (form of a kanji)
- かくにん / 確認 / confirmation
- せつめい / 説明 / explanation
- でんぶん / 伝聞 / hearsay
- もくてき / 目的 / purpose
- ふかひ / 不可避 / inevitability
- けいご / 敬語 / respectful (polite) speech
- ふくじょし / 副助詞 / adverbial particle

docs/search-configuration/custom-dictionaries.md
```diff
@@ -56,18 +56,61 @@ words
...
-The syntax for the provided text file should follow the CSV format:
+The syntax for the provided text file should follow the Comma-Separated Values (CSV) format:
 
 ```csv
-text, token 1 ... token n, reading 1 ... reading n, part-of-speech tag
+text,tokens,readings,part-of-speech
 ```
 
-For example:
+1. `text` is the compound word or phrase that appears in your content, such as a name
```
jazzsequence, line 67: should we indicate that `text` should only contain one word?

jazzsequence (approved):
This is much better and makes a lot more sense even as a non-native speaker.

Left one question, but it's not a blocker. This looks good to me. 👍

docs/search-configuration/custom-dictionaries.md
roborourke: Could do, that's what I'm referring to with compound. Maybe it'll make more sense once it's been translated but I think for now lets return to this in beta 2 or the RC

PR #122, opened 2020-11-18
Add synonyms and stopwords before icu_folding if present

### LINKED ISSUE #121: Ensure synonym filters are added after normalization filters
Search data in Elasticsearch is analysed with a lowercasing filter however neither WordPress nor ElasticPress lowercase the search term before running a query. This means capitalised matches for search terms are dependent on fuzzy matching.

One example of this 

Steps to reproduce:
1. Add a post with only capitalised versions of a word eg. CAPITAL
1. Add another post with only a lowercase version of the word eg. capital
1. Add synonyms to the search config eg. CAPITAL, primary
1. Search for the word on the front end using different capitalisations eg. Capital, CAPITAL, CaPiTaL etc

Capitalisation should not affect the search results matching.

Acceptance criteria:

We could go a bit further here and add a case-sensitive option for Elasticsearch. Or we could just default to all searches being lowercase.

- [ ] Ensure synonyms filters are added after lowercasing

roborourke:
Fixes #121

`icu_folding` does everything that `icu_normalizer` does except that it also converts accented characters to their base forms. For synonyms we want to avoid case-sensitivity but support properly accented words as written by folks in their given language.

This checks for the presence of `icu_folding` and adds synonyms before it if so, otherwise appending them. It also checks for `icu_normalizer` and ensures stopwords are added directly afterwards if present, otherwise prepending them.

### DIFF
--- a/inc/analyzers.json
+++ b/inc/analyzers.json
@@ -854,9 +854,7 @@
 				"kuromoji_baseform",
 				"ja_pos_filter",
 				"icu_normalizer",
-				"icu_folding",
-				"greek_lowercase_filter",
-				"cjk_width"
+				"icu_folding"
 			],
 			"tokenizer": "kuromoji"
 		},
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -902,14 +902,45 @@ function elasticpress_mapping( array $mapping ) : array {
 			$synonyms,
 			$stopwords
 		);
+
 		// Copy default analyzer to default search.
 		$mapping['settings']['analysis']['analyzer']['default_search'] = $mapping['settings']['analysis']['analyzer']['default'];
-		// Add our custom filters.
-		$mapping['settings']['analysis']['analyzer']['default_search']['filter'] = array_merge(
-			array_keys( $synonyms ),
-			array_keys( $stopwords ),
-			$mapping['settings']['analysis']['analyzer']['default_search']['filter']
-		);
+
+		// Add stopwords after `icu_normalizer` if present, otherwise prepend.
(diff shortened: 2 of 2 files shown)


inc/analyzers.json
```diff
@@ -854,9 +854,7 @@
 				"kuromoji_baseform",
 				"ja_pos_filter",
 				"icu_normalizer",
-				"icu_folding",
-				"greek_lowercase_filter",
-				"cjk_width"
+				"icu_folding"
```
roborourke, line 857: Just to note here it turns out `icu_normalizer` actually does the job of `greek_lowercase_filter` and `cjk_width` already

inc/namespace.php
```diff
@@ -902,14 +902,45 @@ function elasticpress_mapping( array $mapping ) : array {
...
-		$mapping['settings']['analysis']['analyzer']['default_search']['filter'] = array_merge(
-			array_keys( $synonyms ),
-			array_keys( $stopwords ),
-			$mapping['settings']['analysis']['analyzer']['default_search']['filter']
-		);
+
+		// Add stopwords after `icu_normalizer` if present, otherwise prepend.
+		// This ensures text is lowercased and full-width characters converted to
+		// half-width while still retaining accents. This is should also ensure
+		// stopwords and stemming have not yet been applied.
```
jazzsequence, line 912: Just a nit, but for multiline comments I would use

```
/*
 * Some stuff here
 */
```
...instead of `//` syntax.
🔢

jazzsequence (approved):

PR #149, opened 2021-03-05
Fix user and term query behaviour

### LINKED ISSUE #140: Make users and terms indexables configurable
The users and terms indexables are active by default and can't be controlled via the Altis config. We should add this in as an option.

### LINKED ISSUE #148: User and term search not working as advertised
Some of our features and some unfortunately unpatched ElasticPress bugs mean that user and term searches aren't behaving properly.

- User search does not work with partial matches
- Autosuggest with meta fields is broken
- Term search using EP is broken due to missing updates from the ES < 7 mapping

roborourke:
There were a number of bugs in ElasticPress meaning that user and term searching was not behaving quite how we would like.

- `sortable` field mapping missing for ES < 7 on some text fields
- search fields using `meta.<key>` without the trailing `.value` property do not work

This fixes #148 and also fixes #140

This should not be backported as there is a breaking change to the ES mapping as a result of fixing this bug.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -93,8 +93,8 @@ The following options can be enabled/disabled via the search configuration.
 - `"facets": true|false|['match-type' => "all" (default)|...]`
 - `"woocommerce": true|false (default)`
 - `"autosuggest": true|false (default)`
-- `"users": (default) true|false`
-- `"terms": (default) true|false`
+- `"users": true (default)|false`
+- `"terms": true (default)|false`
 
 ### Related Posts
 To find related posts leveraging Elastic Search use the `ep_find_related()` function. The function requires a single parameter ( `$post_id` ) with another optional parameter ( `$return` ). The `$post_id` will be used to find the posts that are related to it, with `$return` specifying the number of related posts to return, which defaults to 5.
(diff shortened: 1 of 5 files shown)


docs/README.md
```diff
@@ -93,6 +93,8 @@ The following options can be enabled/disabled via the search configuration.
 - `"facets": true|false|['match-type' => "all" (default)|...]`
 - `"woocommerce": true|false (default)`
 - `"autosuggest": true|false (default)`
+- `"users": (default) true|false`
+- `"terms": (default) true|false`
```
jazzsequence, line 97: Is the `(default)` in the position that it is to indicate that `true` is the default? It looks a little weird when all the other instances have `(default)` placed at the end of the line. Maybe `true (default)|false`?

inc/namespace.php
```diff
@@ -977,34 +988,121 @@ function elasticpress_mapping( array $mapping ) : array {
...
+			$end_field = $sub_fields[ count( $sub_fields ) - 1 ];
+			foreach ( ( $mapping['mappings'][ $type ]['dynamic_templates'] ?? [] ) as $template ) {
+				$template = wp_parse_args( $template[ array_keys( $template )[0] ], [
+					'match_pattern' => null,
+					'match' => '',
+					'unmatch' => '',
+					'path_match' => '',
+					'path_unmatch' => '',
+				] );
+				if ( $template['match_pattern'] === 'regex' && ! preg_match( "/{$template['match']}/", $end_field ) ) {
```
jazzsequence, line 1017: These would benefit from some inline comments to explain what's happening in each of these breaks.

🔢

jazzsequence (approved):
Some very minor comments. Otherwise looks good to me. 👍

PR #174, opened 2021-03-25
Fix ES fallback no-op

joehoyle:
Currently the no-op failed ES queries is not working. This is because the old method of hooking in to `posts_request` filter, checking for `elasticsearch_success`, and returning an empty SQL query string (or well, a SELECT that doesn't match anything). At some point (maybe EP 3), ElasticPress now hooks into `posts_pre_query` which is _after_ `posts_request`, in which case the `elasticsearch_success` is never present.

A better approach is to hook into `posts_pre_query` just after ElasticPress and set the matches posts to an empty array if the EP request just failed.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -306,7 +306,7 @@ function log_remote_request_errors( array $request, ?string $type = null ) {
  * Default ElasticPress functionality is to fall-back to MySQL search when queries fail. We want to instead
  * no-op the query when this happens, as we don't want to put lots of load on to MySQL.
  *
- * @param array $posts
+ * @param ?array $posts List of posts or null if request failed.
  * @param WP_Query $query The current query object.
  * @return array
  */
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -304,31 +306,16 @@ function log_remote_request_errors( array $request, ?string $type = null ) {
  * Default ElasticPress functionality is to fall-back to MySQL search when queries fail. We want to instead
  * no-op the query when this happens, as we don't want to put lots of load on to MySQL.
  *
- * @param string $request SQL query string.
+ * @param array $posts
```
roborourke, line 309: Missing a parameter comment here.
```suggestion
 * @param ?array $posts List of posts or null if request failed.
```

roborourke (approved):
One missing param doc, added that in for you. Guess we'll need to backport to v4 then, that's when EPv3 was introduced.

Just FYI Altis v7 has been updated to EP v3.5.5

PR #185, opened 2021-04-14
Add autosuggest endpoint handling

### LINKED ISSUE #184: The autosuggest configuration option does nothing
Steps to reproduce:

1. [As per the documentation](https://www.altis-dxp.com/resources/developer-docs/search/#search-form-auto-suggest), enable autosuggest in composer.json:

       "modules": {
           "search": {
               "autosuggest": true
           }
       }

2. Publish a page containing a search block
3. Type into the search input and observe that nothing happens

Checking the Network tab of the developer tools I can see that no Ajax requests are being made as I type. Reindexing the site has no effect.

I've also added a search input to the template using `get_search_form()` and this form does not get autosuggest functionality either.

rmccue:
Fixes #184.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1755,14 +1755,14 @@ function handle_autosuggest_endpoint() {
 	/**
 	 * Features instance.
 	 *
-	 * @var Features $features 
+	 * @var Features $features
 	 */
 	$features = Features::factory();
 
 	/**
 	 * Search feature instance.
 	 *
-	 * @var Feature\Search\Search $search 
+	 * @var Feature\Search\Search $search
 	 */
 	$search = $features->get_registered_feature( 'search' );
 
@@ -1792,7 +1792,7 @@ function handle_autosuggest_endpoint() {
 	$client = Elasticsearch::factory();
 
 	// Pass to EP.
-	$response = $client->remote_request( ep_get_index_name() . '/post/_search', [
+	$response = $client->remote_request( Indexables::factory()->get( 'post' )->get_index_name() . '/_search', [
 		'body'   => json_encode( $json ),
 		'method' => 'POST',
 	] );
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -20,6 +20,7 @@ use Psr\Http\Message\RequestInterface;
 use WP_CLI;
 use WP_Error;
 use WP_Query;
+use WP_REST_Server;
 use WP_Term_Query;
 use WP_User_Query;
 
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
 
 	return $request;
 }
+
+/**
+ * Handle request forwarding to ES.
+ */
+function handle_autosuggest_endpoint() {
```
rmccue, line 1732: @roborourke could this actually just be a regular REST API call?

roborourke: That'd need to be handled upstream but presumably for elasticpress.io interop the answer is no. The plugin's JS is where the request is made so would need to modify that.

inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
...
+		wp_send_json( [], 200 );
+	}
+
+	// Validate data.
+	$json = json_decode( WP_REST_Server::get_raw_data(), true );
+	if ( ! $json ) {
+		wp_send_json( [], 200 );
+	}
+
+	/** @var Features $features Features instance. */
```
roborourke, line 1755:
```suggestion
	/**
	 * Features instance.
	 *
	 * @var Features $features 
	 */
```

inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
...
+	$features = Features::factory();
+
+	/** @var Feature\Search\Search $search Search feature instance. */
```
roborourke, line 1758:
```suggestion
	/**
	 * Search feature instance.
	 *
	 * @var Feature\Search\Search $search 
	 */
```

inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
...
+				[
+					'terms' => [
+						'post_type.raw' => array_values( $search->get_searchable_post_types() ),
+					],
+				],
+			],
+		],
+	];
+
+	/** @var Elasticsearch $client */
```
roborourke, line 1779:
```suggestion
	/**
	 * Elasticsearch client object.
	 *
	 * @var Elasticsearch $client
	 */
```

inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
...
+	$client = Elasticsearch::factory();
+
+	// Pass to EP.
+	$response = $client->remote_request( ep_get_index_name() . '/post/_search', [
```
roborourke, line 1783: We should get this directly from the indexable object now instead, `ep_get_index_name()` no longer exists.

inc/namespace.php
```diff
@@ -1717,3 +1725,69 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
...
+	}
+
+	/** @var Features $features Features instance. */
+	$features = Features::factory();
+
+	/** @var Feature\Search\Search $search Search feature instance. */
+	$search = $features->get_registered_feature( 'search' );
+
+	// Force post filter value.
+	$json['post_filter'] = [
```
roborourke, line 1762: Need to double check the requests it makes again to see if there's an existing post filter. I really wish it didn't submit an elasticsearch query straight from JS tbh, would be better to use the actual site search but this is because it was made for elasticpress.io.

roborourke (approved):

PR #191, opened 2021-04-16
Fix ES index and sync healthchecks

roborourke:


inc/namespace.php
```diff
@@ -31,7 +31,10 @@
  */
 function bootstrap() {
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\load_elasticpress', 4 );
-	add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
+	add_action( 'plugins_loaded', function () {
+		// Run healthchecks on indexes after ElasticPress is loaded.
+		add_filter( 'altis_healthchecks', __NAMESPACE__ . '\\add_elasticsearch_healthcheck' );
```
rmccue, line 36: This change doesn't make any sense to me; why would we defer adding a filter?

roborourke: Because ElasticPress isn't loaded yet and it uses ElasticPress code to perform the healthchecks

jazzsequence, line 36: Could the filter be moved into the `load_elasticpress` function or is that too early (because the priority is < 10).

roborourke: I could test it again, it didn't seem to work when I tried it though, I think ElasticPress does some set up on `plugins_loaded` too though I didn't confirm that.

jazzsequence (approved):

PR #200, opened 2021-04-19
Fix `--per-page` doc parameter

igmoweb:
The parameter name and syntax was wrong


roborourke (approved):
Been meaning to do this! Thanks man

PR #203, opened 2021-04-22
Check package file exists before using in mapping

### LINKED ISSUE #181: Synonyms not working after DB import between environments
After importing our production DB into our staging environment, indexing fails until we re-add synonyms in the manual entry meta box.

The same synonyms were configured on both the staging and production environments and after the DB import the correct synonyms were present in the manual entry meta box with the status 'active'.

However, we run into the error in the screenshot below.

image

This also happens when setting up a new local environment with a new DB import.

Steps to reproduce:
1. Import DB from one environment to another.
2. Initiate an elasticpress index.

roborourke:
After importing a database or similar there can be references to non-existent packages which will break the mapping.

Fixes #181


jazzsequence (approved):
👍

PR #213, opened 2021-05-05
Use defined autosuggest endpoint for URI check

### LINKED ISSUE #212: Autosuggest endpoint only works on main site on the network
Following on from #184. The autosuggest feature does not work on subdirectory sites on the network because the `/autosuggest` endpoint returns a 404

Steps to reproduce:

1. Enable `autosuggest` in the search config:

       "modules": {
           "search": {
               "autosuggest": true
           }
       }
2. Create a new site on the network, eg `example.com/site1`
3. Add a search form to a page
4. Try searching in the form and observe there are no autosuggest results

The problem is the `example.com/site1/autosuggest` endpoint returns a 404.

roborourke:
Fixes #212

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1738,13 +1738,25 @@ function split_large_ep_request( $request, array $query, array $args, int $failu
  * Handle request forwarding to ES.
  */
 function handle_autosuggest_endpoint() {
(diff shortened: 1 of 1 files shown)


johnbillion (approved):
Tested and working locally 👍

roborourke:
Much appreciated

PR #221, opened 2021-05-07
Combine equivalent weighted queries

### LINKED ISSUE #220: Combine per post type ES queries that are the same
ElasticPress generates queries that search against each post type individually. This is useful for supporting different field weighting and potentially different treatment of date decay etc but with a lot of post types results in excessively slow queries.

Where the queries match they should be combined into a single query against multiple post types.

Relevant code is in the `do_weighting()` method of the `Weighting.php` class in ElasticPress.

roborourke:
EP creates a search query for each post type which can result in incredibly complex and slow queries, even if the fields to be searched are the same. This will combine equivalent queries.

Fixes #220

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -118,6 +118,9 @@ function load_elasticpress() {
 	// Fix the mime type search query.
 	add_filter( 'ep_formatted_args', __NAMESPACE__ . '\\fix_mime_type_query', 11, 2 );
 
+	// Combine matching post type weighting queries. Weighting added on priority 20 by EP.
+	add_filter( 'ep_formatted_args', __NAMESPACE__ . '\\combine_weighting_queries', 21 );
+
 	// Back compat for ElasticPress v2 - change post index name to old version.
 	add_filter( 'ep_index_name', __NAMESPACE__ . '\\filter_index_name' );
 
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -1479,6 +1534,7 @@ function get_autosuggest_fields( ?string $type = null ) : array {
 	// Add default autosuggest fields.
 	if ( isset( $fields['post'] ) ) {
 		$fields['post'] = [ 'post_title' ];
+		$fields['post'] = [ 'post_name' ];
```
kadamwhite, line 1537: Not sure I understand this change -- we're overwriting the value here, should we be setting this to a list of `post_title` _and_ `post_name`? What's the intent of this line?

kadamwhite (changes requested):
Ah, actually, it's more complex than that. Using a sample database of news stories and searching for a company name, for example "Warner Bros," without this change I get about 60 results, all of which contain "Warner" and "Bros". Now, with this change, the search takes approximately the same amount of user-facing time, but now I am receiving _205_ pages of results, many of which include no mention of _either_ word. The logic here is wrong somehow I think.

inc/namespace.php
roborourke: Oh whoops, that was some other work in progress unrelated to this PR, I'll remove it.

roborourke:
Sorry @kadamwhite, there was an unrelated change that I've removed, it could potentially have affected the results you were getting so I'd try it again now. Would you also be able to paste the generated query here? I'll get a proper test together for myself locally too.

roborourke:
Ok, so traced it down to this code that will be different between Altis v6 and v7:

https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L1258-L1273

In Altis v7 ElasticPress has been updated which includes an updated search algorithm that essentially removes the loose phrase match that resulted in a long tail of irrelevant search results. Altis removes this already but with a checkout of this branch (i.e. based on the current `master`) the search algorithm version check would fail and the loose match would not be removed. This is why you got such different results.

As you noted rebasing these commits on the `v6-branch` shows no difference.

kadamwhite (approved):
Acknowledging @roborourke's note above, this change looks good and is doing what I'd expect locally. Back to :heavy_check_mark:, thank you for investigating!

jazzsequence (commented):
Can we put this on a list for needed unit tests? Can this be unit tested? Seems like complex search stuff should be one of the things we're testing heavily.

PR #222, opened 2021-05-07
Limit search queries to 100 characters

roborourke:
This one will need a little discussion I think on what a reasonable limit is.

This hopefully will prevent spam queries tying up resources for too long with absurdly long search queries.

### DIFF
--- a/docs/search-configuration/README.md
+++ b/docs/search-configuration/README.md
@@ -141,7 +141,7 @@ This option allows you to prevent transpositions from being counted as a single 
 
 ## Max Query Length
 
-Typically search queries can be as long as the user wishes. As a preventative measure against slow queries and search request spamming Altis limits search query strings to **80 characters** by default.
+Typically search queries can be as long as the user wishes. As a preventative measure against very slow queries and search request spamming Altis limits search query strings to **100 characters** by default.
 
 You can configure this value in 2 ways.
 
@@ -153,7 +153,7 @@ Using the Altis Search config option `max-query-length`:
 		"altis": {
 			"modules": {
 				"search": {
-					"max-query-length": 100
+					"max-query-length": 150
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


ntwb:
URLs would be the use case that comes to mind that this would have an impact on, one project I know of has both a long domain and sub-domain before even getting to the path in the URL and they perform a lot of searches for pages using the page URL.

docs/search-configuration/README.md
```diff
@@ -138,3 +138,35 @@ If you have a prefix length of 0 you may wish to increase this value to get more
 **`transpositions`**
 
 This option allows you to prevent transpositions from being counted as a single edit. In the above example the transposition "act" to "cat" has an edit distance of 1. Setting this value to `false` would mean the same transposition would have an edit distance of 2 because 2 letters have been replaced.
+
+## Max Query Length
+
+Typically search queries can be as long as the user wishes. As a preventative measure against slow queries and search request spamming Altis limits search query strings to **80 characters** by default.
```
jazzsequence, line 144:
```suggestion
Typically search queries can be as long as the user wishes. As a preventative measure against slow queries and search request spamming, Altis limits search query strings to **80 characters** by default.
```

inc/namespace.php
```diff
@@ -1823,3 +1826,33 @@ function handle_autosuggest_endpoint() {
...
+
+	// Get max length from config.
+	$max_length = Altis\get_config()['search']['max-query-length'] ?? 80;
+
+	/**
+	 * Filters the maximum allowed query string length for searches.
+	 *
+	 * @param int $max_length The maximum allowed query string length.
+	 */
+	$max_length = (int) apply_filters( 'altis.search.max_query_length', $max_length );
```
jazzsequence, line 1852: To @ntwb's point, we could pass the `$search` value into the filter here, so you could filter this based on the search itself, e.g. if the `$search` looked like a URL, update the `$max_length` to some suitable large number, otherwise keep the default behavior.

```suggestion
	$max_length = (int) apply_filters( 'altis.search.max_query_length', $max_length, $search );
```

jazzsequence (commented):
I'm happy with this in that it can be filtered or changed from the config. I want to be cognizant of other use cases, including @ntwb's (which baffles me personally from a user perspective, but seems like something we should be conscious of).

PR #226, opened 2021-06-01
Update searchable fields documentation

roborourke:
The searchable fields docs are out of date since Altis v5 when ElasticPress added weighting. The `ep_search_fields` filter only works for queries where specific fields have been provided.

### DIFF
--- a/docs/cms-query-integration.md
+++ b/docs/cms-query-integration.md
@@ -37,6 +37,70 @@ add_action( 'pre_get_posts', function ( WP_Query $query ) {
 }, 20 );
 ```
 
+## Customising Searched Fields
+
+Altis will search key content fields by default, including those defined in the [field boosting config](./search-configuration/README.md#field-boosting), depending on the type of content.
+
+### Using Query Parameters
+
+`WP_Query`, `WP_Term_Query` and `WP_User_Query` support a `search_fields` parameter. This can be used override the default searched fields and even to modify how the fields are boosted using the `<field>^<boost value>` syntax.
+
+```php
+$posts = new WP_Query( [
+	'ep_integrate' => true, // Use Elasticsearch.
+	'post_type' => 'project',
+	'search_fields' => [
+		'post_title',
+		'meta.client.value^2',
+	],
+] );
+```
+
+Note that when using the `search_fields` parameter the `ep_search_fields` filter will be run and can modify the end results.
+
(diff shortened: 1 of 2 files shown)


docs/cms-query-integration.md
```diff
@@ -37,6 +37,70 @@ add_action( 'pre_get_posts', function ( WP_Query $query ) {
 }, 20 );
 ```
 
+## Customising Searched Fields
+
+Altis will search key content fields by default, including those defined in the [field boosting config](./search-configuration/README.md#field-boosting), depending on the type of content.
+
+### Using Query Parameters
+
+`WP_Query`, `WP_Term_Query` and `WP_User_Query` support a `search_fields` parameter. This can be used override the default searched fields and even to modify how the fields are boosted using the `<field>^<boost value>` syntax.
```
jazzsequence, line 46: Missing word

```suggestion
`WP_Query`, `WP_Term_Query` and `WP_User_Query` support a `search_fields` parameter. This can be used to override the default searched fields and even to modify how the fields are boosted using the `<field>^<boost value>` syntax.
```

docs/cms-query-integration.md
```diff
@@ -37,6 +37,70 @@ add_action( 'pre_get_posts', function ( WP_Query $query ) {
...
+	'ep_integrate' => true, // Use Elasticsearch.
+	'post_type' => 'project',
+	'search_fields' => [
+		'post_title',
+		'meta.client.value^2',
+	],
+] );
+```
+
+Note that when using the `search_fields` parameter the `ep_search_fields` filter will be run and can modify the end results.
```
jazzsequence, line 59: I think we can just remove the "Note that" here. Also wondering if we need to use the phrase "end results" specifically? I only say this because it sounds weird saying out loud (or in my head) right after another "and", and "modify the results" makes just as much sense. (If not more....we're talking about search results, right?)

```suggestion
When using the `search_fields` parameter, the `ep_search_fields` filter will be run and can modify the results.
```

jazzsequence (changes requested):
Some minor changes to make it sound clearer. Otherwise lgtm 👍

docs/cms-query-integration.md
roborourke: derp

jazzsequence (approved):
👍  ship it  🚀

PR #261, opened 2021-07-27
[docs] Add a note about Elastic results availability during (re-)index

shadyvb:
This adds a note about the unavailability of Elastic during (re-)index processes, eg: faceting, autosuggest, and the query integration.

### DIFF
--- a/docs/reindexing-content.md
+++ b/docs/reindexing-content.md
@@ -14,6 +14,8 @@ wp site list --field=url | xargs -I % wp elasticpress index --url=%
 
 `wp site list --field=url` will print each site URL on a new line and ingested by `xargs` to be processed individually.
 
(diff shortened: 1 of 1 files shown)


docs/reindexing-content.md
```diff
@@ -14,6 +14,8 @@ wp site list --field=url | xargs -I % wp elasticpress index --url=%
 
 `wp site list --field=url` will print each site URL on a new line and ingested by `xargs` to be processed individually.
 
+**Note:** During the indexing process, the site will not be using Elastic index for searches/queries, so some features like faceting or autosuggest will not work as expected until the reindex process is finished.
```
rmccue, line 17:
```suggestion
**Note:** During the indexing process, the site will not use the Elasticsearch index for searches/queries, so some features like faceting or autosuggest will not work as expected until the process is finished.
```

rmccue (approved):

PR #275, opened 2021-08-20
Fix mapping modification support for ES 7

roborourke:
I missed that in ES 7 the mappings key has no subtype property so our mapping modifications were not being applied.

This change is suitable for EP 3.4+ so backporting to Altis v5.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -24,6 +24,8 @@ use WP_REST_Server;
 use WP_Term_Query;
 use WP_User_Query;
 
+const SORTABLE_MAX_LENGTH = 10922;
+
 /**
  * Bootstrap search module.
  *
@@ -1131,7 +1133,7 @@ function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
 			$new_mapping['properties'][ $name ]['fields'] = array_merge_recursive_distinct( $property['fields'], [
 				'sortable' => [
 					'type'         => 'keyword',
-					'ignore_above' => 10922,
+					'ignore_above' => SORTABLE_MAX_LENGTH,
 					'normalizer'   => 'lowerasciinormalizer',
 				],
 			] );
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -833,9 +833,21 @@ function elasticpress_analyzer_language() : string {
...
  *
  * @param array $mapping Mapping array.
+ * @param string|null $index The current index this mapping is for.
  * @return array
  */
-function elasticpress_mapping( array $mapping ) : array {
+function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
+	// Derive current mapping type.
+	$indexables = Indexables::factory()->get_all( null, true );
+	preg_match( '/.+-(' . implode( '|', $indexables ) . ')-\d+$/', $index, $matches );
```
rmccue, line 842: This should be `preg_quote`'d just in case
```suggestion
	$escaped_indexables = array_map( function ( $item ) {
		return preg_quote( $item, '/' );
	}, $indexables );
	preg_match( '/.+-(' . implode( '|', $escaped_indexables ) . ')-\d+$/', $index, $matches );
```

inc/namespace.php
```diff
@@ -1006,128 +1018,129 @@ function elasticpress_mapping( array $mapping ) : array {
...
-						'normalizer'   => 'lowerasciinormalizer',
-					],
-				] );
-			}
+	foreach ( $new_mapping['properties'] as $name => $property ) {
+		if ( $property['type'] === 'text' && isset( $property['fields'] ) ) {
+			$new_mapping['properties'][ $name ]['fields'] = array_merge_recursive_distinct( $property['fields'], [
+				'sortable' => [
+					'type'         => 'keyword',
+					'ignore_above' => 10922,
```
rmccue, line 1131: Could we extract this magic number into a descriptive named constant?

rmccue (changes requested):

inc/namespace.php
roborourke: It's just matching the default ElasticPress mapping for sortable fields. We could have a constant but it'd only be used this once.

rmccue: Yeah, I would still move to a constant

rmccue (approved):

PR #288, opened 2021-09-07
Default to inline index settings

shadyvb:
- Defaults to inline index settings, configurable through Altis config.
- Adds textarea input for Japanese user dictionary, and aligns the code for the three package types.
- Updates docs to reference the updates and config option.

### DIFF
--- a/docs/search-configuration/custom-dictionaries.md
+++ b/docs/search-configuration/custom-dictionaries.md
@@ -10,7 +10,9 @@ User dictionaries can be configured for [the entire network](admin://network/adm
 
 Note that for the user dictionary, manual entries will override any uploaded files. So make sure to use one or the other, not both, unlike what you can do with synonyms and stop words.
 
-By default, Altis uses inline index settings to include all of these dictionaries, however, it is recommended to turn such feature off in case uploaded files are bigger than 100KB for the sake of performance and ES cluster sizes ( see below ). Altis will notify you if your file size exceeds the recommended limit within the configuration page.
+By default, Altis uses inline index settings to include all of these dictionaries, however, it is recommended to turn this feature off in case uploaded files are bigger than 100KB. This helps to improve the performance and ES cluster sizes (see below). Altis will notify you if your file size exceeds the recommended limit within the configuration page.
+
+**Note**: If you do not use inline index settings you will need to manually reindex your content after making changes to synonyms, stopwords or the Japanese user dictionary.
 
 ```json
 {
--- a/docs/search-configuration/custom-dictionaries.md
+++ b/docs/search-configuration/custom-dictionaries.md
@@ -12,7 +12,7 @@ Note that for the user dictionary, manual entries will override any uploaded fil
 
 By default, Altis uses inline index settings to include all of these dictionaries, however, it is recommended to turn such feature off in case uploaded files are bigger than 100KB for the sake of performance and ES cluster sizes ( see below ). Altis will notify you if your file size exceeds the recommended limit within the configuration page.
 
-```
+```json
 {
     "extra": {
         "altis": {
--- a/docs/search-configuration/custom-dictionaries.md
+++ b/docs/search-configuration/custom-dictionaries.md
@@ -2,11 +2,29 @@
 
 The search module supports 3 types of custom dictionaries for helping to tune search results for your content. These are synonyms, stop words and custom text analysis for Japanese.
 
-In the website admin, you can upload text files (`.txt` only) and, in the case of stop words and synonyms, also manually enter them on the [Search Configuration](admin://admin.php?page=search-config) screen.
+In the website admin, you can upload text files (`.txt` only) and/or manually enter them on the [Search Configuration](admin://admin.php?page=search-config) screen.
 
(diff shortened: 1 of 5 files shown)


inc/namespace.php
```diff
@@ -1151,6 +1187,19 @@ function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
...
 
+/**
+ * Returns where inline settings should be used.
+ *
+ * @return bool
+ */
+function should_inline_settings() : bool {
+	$es_version = Elasticsearch::factory()->get_elasticsearch_version();
+
+	return version_compare( $es_version, '7.4', '>=' )
```
roborourke, line 1198: Can we add a comment on why it's version 7.4 here? e.g.
```suggestion
	// AWS Elasticsearch 7.4+ is required for the _open and _close API to update index setting.
	// See https://docs.aws.amazon.com/opensearch-service/latest/developerguide/supported-operations.html#version_api_notes
	return version_compare( $es_version, '7.4', '>=' )
```

inc/namespace.php
```diff
@@ -924,60 +926,94 @@ function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
...
-					$stopwords[ "{$sub_type}_{$type}_filter" ] = [
-						'type' => 'stop',
-						'ignore_case' => true,
-						'stopwords_path' => $package_id,
-					];
+					if ( $inline_index_settings ) {
+						$stopwords[ "{$sub_type}_{$type}_filter" ] = [
+							'type' => 'stop',
+							'ignore_case' => true,
+							'stopwords' => implode( ',', $package_contents ),
```
roborourke, line 986: Any reason not to pass this as an array as well? Should be supported iirc.

roborourke (changes requested):
Some minor things to change but really appreciate the additional rationalisation of the existing implementation!

inc/namespace.php
shadyvb: True, the docs just referenced a string so i went with that, but looking again it does indeed support arrays as well.

shadyvb:
```suggestion
							'stopwords' => $package_contents,
```

inc/namespace.php
shadyvb: @roborourke I haven't actually touched the index update logic, does it happen behind the scenes ? or did I miss a whole block of functionality to cover ?

inc/packages/templates/config.php
```diff
@@ -5,6 +5,10 @@
  * @package altis/search
  */
 
+use const Altis\Enhanced_Search\Packages\MAX_INLINE_SETTINGS_SIZE_RECOMMENDED;
```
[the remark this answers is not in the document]
shadyvb: I used FQDN versions as it seems we cannot use the parent namespace in template files.

inc/namespace.php
shadyvb: @roborourke Also, the 7.4 check is for the user dictionary changes, which only came in effect after 7.4.

shadyvb: This is more like the actual reason:

```suggestion
	// Elasticsearch 7.4 is where the inline user_dictionary_rules config option was introduced.
	return version_compare( $es_version, '7.4', '>=' )
```

docs/search-configuration/custom-dictionaries.md
```diff
@@ -2,12 +2,30 @@
...
+Altis includes support for a Japanese user dictionary to be used by the [kuromoji tokenizer](https://www.elastic.co/guide/en/elasticsearch/plugins/6.3/analysis-kuromoji-tokenizer.html) that splits words up during analysis. Synonyms and stop words are "filters" so multiple files (uploaded and manually entered) can be used for those.
 
 User dictionaries can be configured for [the entire network](admin://network/admin.php?page=search-config) (provided the subsites match the primary site language) or at an individual site level.
 
+Note that for the user dictionary, manual entries will override any uploaded files. So make sure to use one or the other, not both, unlike what you can do with synonyms and stop words.
+
+By default, Altis uses inline index settings to include all of these dictionaries, however, it is recommended to turn such feature off in case uploaded files are bigger than 100KB for the sake of performance and ES cluster sizes ( see below ). Altis will notify you if your file size exceeds the recommended limit within the configuration page.
```
roborourke, line 13:
```suggestion
By default, Altis uses inline index settings to include all of these dictionaries, however, it is recommended to turn this feature off in case uploaded files are bigger than 100KB. This helps to improve the performance and ES cluster sizes (see below). Altis will notify you if your file size exceeds the recommended limit within the configuration page.

**Note**: If you do not use inline index settings you will need to manually reindex your content after making changes to synonyms, stopwords or the Japanese user dictionary.
```

roborourke (approved):
Good work @shadyvb!

I realised there will need to be a follow up on the Cloud module too to check this inline index settings option and the ES version to avoid the AWS packages integration stuff.

shadyvb:
@roborourke I thought about that, but decided to keep it in place so the switching happens on the software side, rather than cloud. Can't see a downside to keeping it uploading packages, so if the switch happens here ( due to filesize limits, etc ) the package will already be stored in Cloud. 

Does that make sense ?

roborourke:
Oh yeah - that component of it is fine and makes sense, no issue there. This PR is totally fine.

There are 2 steps to the cloud integration for dictionary files on AWS:

- Save the dictionary content to S3
- Create an Elasticsearch "Package" via the AWS API using the file on S3

The second part is done in the cloud module.

Issue #289, opened 2021-09-13
Fix failure to remove/clear Elastic packages


shadyvb:
Steps to reproduce:
1. Go to Search config admin page
2. Add uploaded or manual packages to any of the three package types ( synonyms, stopwords, user dictionary )
3. Save
4. Try removing/clearing the added package.

Expected: the package is cleared/removed
Actual: the package is not cleared/removed

garethwelton:
Hei @shadyvb we don't tend to put story point values onto bugs, I'll add some comments to this is the playbook if not already meantioned.

roborourke:
Fixed by #294

PR #294, opened 2021-09-27
Fix clearing search packages

shadyvb:
- Removes packages when passed empty content.
- Fixes a fatal with update index call for each package deletion, which fatals but silently since it is a cron job.

humanmade/altis-enhanced-search#289

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -533,9 +533,6 @@ function delete_package( string $slug, bool $for_network = false ) {
 		);
 	}
 
-	// Update the index settings on successful deletion.
-	do_settings_update( $for_network, strpos( $slug, 'user-dictionary' ) !== false );
-
 	/**
 	 * Action triggered when a package is deleted.
 	 *
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -320,7 +320,7 @@ function handle_form() : void {
 		$package_id = null;
 
 		// Handle manual entry.
-		if ( ! empty( $_POST[ $text_field ] ) ) {
+		if ( isset( $_POST[ $text_field ] ) ) {
 			$text = sanitize_textarea_field( wp_unslash( $_POST[ $text_field ] ) );
 			$file = get_package_path( "manual-{$type}", $for_network );
 			$has_changed = true;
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -524,9 +533,6 @@ function delete_package( string $slug, bool $for_network = false ) {
 		);
 	}
 
-	// Update the index settings on successful deletion.
-	do_settings_update( $for_network, strpos( $slug, 'user-dictionary' ) !== false );
```
roborourke, line 528: Do you know any more about what the fatal is or why it happens? How are indexes updated if a package is deleted now?

shadyvb: The fatal was happening because [`update_index_settings`](https://github.com/humanmade/altis-enhanced-search/blob/6eab34da0c9ebcc176dc70fa2e3b07a52e975708/inc/packages/namespace.php#L606) called from [`do_settings_update`](https://github.com/humanmade/altis-enhanced-search/blob/6eab34da0c9ebcc176dc70fa2e3b07a52e975708/inc/packages/namespace.php#L589) expects `settings` to not be null, which it was always being passed as null, unsure really why is that happening tbh, didn't quite get the logic in [`elasticpress_mapping`](https://github.com/humanmade/altis-enhanced-search/blob/6eab34da0c9ebcc176dc70fa2e3b07a52e975708/inc/namespace.php#L843) that returns early, but it certainly fatal'd all the time.

And re indexes being updated, that's happening in [`handle_form`](https://github.com/humanmade/altis-enhanced-search/blob/6eab34da0c9ebcc176dc70fa2e3b07a52e975708/inc/packages/namespace.php#L398) where `delete_package` is called (only place I could see it being used). `create_package` does not directly update indexes as well, so seemed like `delete_package` should also be consistent around that, unless I missed something that isn't obvious. IMHO, delete package should not have more side-effects than just deleting the package, consumers should update indexes separately (specially if any bulk operation is happening, otherwise multiple settings update calls can happen).

roborourke: Aha yeah, it's covered by this hook: `add_action( 'altis.search.updated_packages', __NAMESPACE__ . '\\on_updated_packages', 10, 2 );`. Maybe it was something I forgot to remove.

In the cloud module however, we remove that hook in favour of waiting until the packages have been created and associated with the AWS ES domain.

I think we need to take another look at that. Backwards compatibility is useful but there's no need to wait for the package association steps if we're adding the settings inline.

roborourke (approved):

PR #309, opened 2021-11-02
Add filter to autosuggest endpoint to modify query

### LINKED ISSUE #308: Autosuggest - cant filter the results as `post_filter` is hardcoded.
My use case: I want to prevent posts with the meta key 'unlisted' from being displayed in autosuggest results. 

I have tried filtering the query used to pass either a `meta_query`, or `post__not_in`, but neither approach works. In Chrome dev tools I can see the correct es query options being passed as part of the request to `/autosuggest`, but these seem to be having no effect. 

Further investigation it is because this is using the`post_filter` argument, and this is completely overridden here: https://github.com/humanmade/altis-enhanced-search/blob/master/inc/namespace.php#L1981

IMO this is a bug, as I would expect this to work. It is also pretty opaque that this argument is hardcoded and took me a couple of hours to work out why this didn't work. 

Not opening a PR as I don't really have a deep knowledge of this functionality, but perhaps this would be a better solution: 

```php
	// Force post filter value.
	$json = array_merge_recursive( $json, [
		'post_filter' => [
			'bool' => [
				'must' => [
					[
						'term' => [
							'post_status' => 'publish',
						],
					],
					[
						'terms' => [
							'post_type.raw' => array_values( $search->get_searchable_post_types() ),
						],
					],
				],
			],
			],
	] );
```

mattheu:
My use case is that I want to restrict what is shown in the autosuggest endpoint, to hide content with a certain meta key. 

A filter on the query would be useful to do this. It is a bit of a tricky one as the format is an ES query, not a WP query. 

Fixes #308

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1978,8 +1978,8 @@ function handle_autosuggest_endpoint() {
 	$search = $features->get_registered_feature( 'search' );
 
 	// Force post filter value.
-	// This is necessary because the ES query is passed frmo the client side,
-	// and we want to ensure only published, searchable content.
+	// This is necessary because the ES query is passed from the client side,
+	// and we want to ensure only published, searchable content is returned.
 	$json['post_filter'] = [
 		'bool' => [
 			'must' => [
@@ -1997,7 +1997,6 @@ function handle_autosuggest_endpoint() {
 		],
 	];
 
-	
 	/**
 	 * Filter the autosuggest query JSON.
 	 *
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1978,6 +1978,8 @@ function handle_autosuggest_endpoint() {
 	$search = $features->get_registered_feature( 'search' );
 
 	// Force post filter value.
+	// This is necessary because the ES query is passed frmo the client side,
+	// and we want to ensure only published, searchable content.
 	$json['post_filter'] = [
 		'bool' => [
 			'must' => [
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -1978,6 +1978,8 @@ function handle_autosuggest_endpoint() {
 	$search = $features->get_registered_feature( 'search' );
 
 	// Force post filter value.
+	// This is necessary because the ES query is passed frmo the client side,
+	// and we want to ensure only published, searchable content.
```
roborourke, line 1982:
```suggestion
	// This is necessary because the ES query is passed from the client side,
	// and we want to ensure only published, searchable content is returned.
```

inc/namespace.php
```diff
@@ -1995,6 +1997,16 @@ function handle_autosuggest_endpoint() {
 		],
 	];
 
+
```
roborourke, line 2000:
```suggestion
```

inc/namespace.php
```diff
@@ -1995,6 +1997,16 @@ function handle_autosuggest_endpoint() {
...
 
+	
+	/**
+	 * Filter the autosuggest query JSON.
+	 *
+	 * This is the data that is json encoded and passed as the elasticsearch request body.
+	 *
+	 * @param array $json ES Query.
+	 */
+	$json = apply_filters( 'altis.search.autosuggest_json', $json );
```
roborourke, line 2008: Only thing I'm wondering here is the filter name - to be honest the variable isn't named that well because it's not JSON yet. I'm thinking `altis.search.autosuggest_query`?

roborourke (changes requested):
Couple of linter errors I've made suggestions for.

One thing missing here is documentation of the filter. I guess under the "Search Form Auto Suggest" on the docs readme will do for now.

That feature could do with more complete docs and some distinction from the `WP_Query` level autosuggest but I'd say that's out of scope here.

roborourke (approved):

PR #311, opened 2021-11-04
Fix elasticpress get-mapping command check

roborourke:
The `find_command_to_run()` method expects an array of positional arguments rather than a string containing the subcommand.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -224,7 +224,7 @@ function configure_documents_feature() {
  */
 function add_elasticpress_get_mapping_subcommand() {
 	// Check if command already exists.
-	if ( is_array( WP_CLI::get_runner()->find_command_to_run( 'elasticpress get-mapping' ) ) ) {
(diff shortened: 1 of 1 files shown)


roborourke:
cc @tfrommen - I'm not sure when the behaviour of this changed I just started seeing the following error and this fixed it! Thought you might find it interesting.

```
Warning: array_shift() expects parameter 1 to be array, string given in phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/RootCommand.php on line 39
```

tfrommen (approved):

tfrommen:
@roborourke sorry for that. 😩 I'm sure I just assumed this to be in line with `wp cli has-command` or the `WP_CLI::runcommand` function, both of which take a string. (I added that after the command itself, and might not have tested this properly.)

roborourke:
No it was working fine, not your fault, it’s just an update to the PHP container in Altis v9 (PHP sessions support via redis) that I think has an updated version of WP CLI

PR #315, opened 2021-11-11
Update dependabot config for v9

roborourke:


.github/dependabot.yml
```diff
@@ -3,35 +3,44 @@ updates:
 - package-ecosystem: composer
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
```
robindevitt, line 6: How come this line gets indented? In some scenarios there is the "-"?

robindevitt: Rob covered the explanation here https://hmn.slack.com/archives/C5WSNGC9F/p1636710824228000

robindevitt (approved):

PR #336, opened 2021-11-25
Prevent memory leak in ElasticPress CLI with debug mode

### LINKED ISSUE #335: Memory leak in elasticpress indexing
When ElasticPress indexes post / other objects, I believe it never frees any memory, so `$wp_object_cache->cache` grows larger and larger. This uses a lot of memory for sites with 100k+ posts, but also causes a significant performance bottleneck as the `->cache` array becomes huge so lookups etc take a long time.

Like many other long running things, something like a local cache flush might be useful. 

For reference, after import 90k posts, the process is using about 1.2GB memory.

roborourke:
Fixes #335

### DIFF
--- a/load.php
+++ b/load.php
@@ -31,3 +31,14 @@ add_action( 'altis.modules.init', function () {
 	];
 	Altis\register_module( 'search', __DIR__, 'Search', $options, __NAMESPACE__ . '\\bootstrap' );
 } );
+
+// Ensure ElasticPress commands don't use debug mode as it can result
+// in out of memory errors on big sites as queries are logged during bulk updates.
+add_action( 'altis.loaded_autoloader', function () {
+	if ( ! defined( 'WP_CLI' ) || ! WP_CLI || ! isset( $_SERVER['argv'] ) ) {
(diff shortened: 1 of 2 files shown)


joehoyle (approved):

load.php
```diff
@@ -31,3 +33,15 @@
...
 } );
+
+// Ensure ElasticPress commands don't use debug mode as it can result
+// in out of memory errors on big sites as queries are logged during bulk updates.
+add_action( 'altis.loaded_autoloader', function () {
+	if ( ! defined( 'WP_CLI' ) || ! WP_CLI || ! isset( $_SERVER['argv'] ) ) {
+		return;
+	}
+	// phpcs:ignore HM.Security.ValidatedSanitizedInput -- Comparing $_SERVER value to a string directly.
+	if ( ( $_SERVER['argv'][1] ?? '' ) === 'elasticpress' ) {
```
rmccue, line 44: `wp --url=foo elasticpress ...` won't work with this afaik? I think there's a WP CLI helper to get this.

PR #341, opened 2021-12-02
Fix array key

tfrommen:
This PR fixes an assumed copy-paste mistake. The code checks for existence of one array element/path, but then accesses another one (the same as the code block before it).

With configurations/queries where `terms` is not set, but `regexp` is, this will trigger two PHP notices.


roborourke (approved):
Thanks!

PR #346, opened 2021-12-03
Add initial QA tests

roborourke:
Fixes https://github.com/humanmade/product-dev/issues/829


shadyvb (approved):

roborourke:
These should be ok even without the other PRs, just ended with those while I was going in circles trying to figure out why things to work inconsistently / intermittently.

PR #347, opened 2021-12-03
Bug fixes for ES user dictionaries

roborourke:
Automatic updates to the index settings when settings synonyms etc was failing during testing, partly caused by the introduction of checking the indexable type in the mapping config filter.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -327,7 +327,7 @@ function handle_form() : void {
 			$has_changed = true;
 
 			// Check for updates.
-			// phpcs:ignore -- Allow error supporession for existing ID check when file exists.
+			// phpcs:ignore -- Allow error suppression for existing ID check when file exists.
 			if ( file_exists( $file ) && ! empty( @get_package_id( "manual-{$type}", $for_network ) ) ) {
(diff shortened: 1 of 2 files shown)


inc/packages/namespace.php
```diff
@@ -326,7 +327,8 @@ function handle_form() : void {
 			$has_changed = true;
 
 			// Check for updates.
-			if ( file_exists( $file ) ) {
+			// phpcs:ignore -- Allow error supporession for existing ID check when file exists.
```
shadyvb, line 330:
```suggestion
			// phpcs:ignore -- Allow error suppression for existing ID check when file exists.
```

inc/packages/namespace.php
```diff
@@ -580,7 +582,7 @@ function get_site_indices( ?int $blog_id = null ) : string {
  */
 function do_settings_update( bool $for_network = false, bool $update_data = false ) : void {
 	// Get latest settings.
-	$mapping = Enhanced_Search\elasticpress_mapping( [] );
+	$mapping = Enhanced_Search\elasticpress_mapping( [], 'x-post-1' );
```
shadyvb, line 585: Can we add a comment here on why we used `x-post-1` ?

shadyvb (changes requested):

shadyvb (approved):

PR #349, opened 2021-12-06
Fix autosuggest mappings

roborourke:
Fixes missing meta property autosuggest mapping support due to forgetting to update a specific variable that deals with dynamic template mappings.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -900,10 +900,11 @@ function elasticpress_analyzer_language() : string {
 function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
 	// Derive current mapping type.
 	$indexables = Indexables::factory()->get_all( null, true );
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -900,10 +900,11 @@ function elasticpress_analyzer_language() : string {
 function elasticpress_mapping( array $mapping, ?string $index = null ) : array {
 	// Derive current mapping type.
 	$indexables = Indexables::factory()->get_all( null, true );
+	$indexables = array_merge( $indexables, Indexables::factory()->get_all( true, true ) );
```
roborourke, line 903: The list excluded `user` because the first parameter for `get_all()` determines whether to get indexables scoped to the site or to the whole network - the user indexable is network wide.

shadyvb (approved):

PR #355, opened 2021-12-13
Index all post taxonomies

### LINKED ISSUE #354: Integrate authorship with ElasticPress author queries
Right now it seems an `author__in` query won't work when ElasticPress kicks in. Need to investigate that and if it's the case make sure that this works seamlessly.

roborourke:
Fixes #354


roborourke:
cc @atimmer

shadyvb (approved):

PR #369, opened 2022-01-19
Add theme for tests

roborourke:
Should hopefully fix the test run, realised there's no default theme with local server and we're checking some front end stuff in these tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+tests/_output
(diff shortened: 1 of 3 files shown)


roborourke:
🚀 🎉 🙌 💥

roborourke:
We might want to consider adding a default theme to `module.yml` in the dev tools repo to keep things consistent and avoid needing to do too much of this kind of hack.

Also moving the `composer server start` to `before_script` would make it easy to use `composer server cli theme install ... -- --activate` for example.

shadyvb (approved):

shadyvb:
I was going to suggest the same. Good thinking.

roborourke:
Rapid :)

PR #383, opened 2022-02-11
Validate uploaded ES package files

shadyvb:
Ref: https://github.com/humanmade/product-dev/issues/885

CleanShot 2022-02-11 at 18 27 26@2x

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -363,23 +363,33 @@ function handle_form() : void {
 
 		// Handle file upload.
 		if ( ! empty( $_FILES ) && isset( $_FILES[ $file_field ] ) && ! empty( $_FILES[ $file_field ]['tmp_name'] ) ) {
-			$file = get_package_path( "uploaded-{$type}", $for_network );
-			// phpcs:ignore HM.Security.ValidatedSanitizedInput.InputNotSanitized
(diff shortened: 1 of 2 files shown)


inc/packages/namespace.php
```diff
@@ -363,23 +363,34 @@ function handle_form() : void {
...
-			if ( ! $result ) {
+			// phpcs:ignore HM.Security.ValidatedSanitizedInput
+			$mime_type = \mime_content_type( $_FILES[ $file_field ]['tmp_name'] );
+			if ( $mime_type !== 'text/plain' ) {
 				$errors[] = new WP_Error(
-					'write_package_error',
+					'file_type_incorrect',
 					// translators: %s replaced by search package file path.
-					sprintf( __( 'Could not write search package file to %s', 'altis' ), $file )
+					sprintf( __( 'Detected unsupported file type %, only textual files are supported.', 'altis' ), $mime_type )
```
roborourke, line 372: Not sure textual is a word, but just text here is fine
```suggestion
					sprintf( __( 'Detected unsupported file type %, only text files are supported.', 'altis' ), $mime_type )
```

roborourke (approved):
Two suggestions that I'll commit, and a code style note for future ref.

PR #388, opened 2022-03-09
Improve Search documentation

rmccue:
I mean, what else are you going to do on a 17 hour flight?


joehoyle (approved):
Looks great, much better indeed

PR #396, opened 2022-03-21
Add interval between bulk indexing requests

roborourke:


joehoyle:
I think we'd be better adding auto-retry / exponential backoff

PR #397, opened 2022-03-21
Update Travis file reference

roborourke:
Ref https://github.com/humanmade/altis-dev-tools/pull/328


roborourke:
Wonder why using twentytwentyone was breaking it 🤔 well, at least tests are passing again 🎉

PR #405, opened 2022-04-28
Unnecessary HTTP API request on pages that enqueue autosuggest

johnbillion:
On a site that uses autosuggest, a server-side HTTP API request to ES is triggered when the autosuggest JavaScript is enqueued. It appears this request is unnecessary.

Steps to reproduce:

1. Enable autosuggest in composer.json
   ```json
   "search": {
   	"autosuggest": true
   },
   ```
2. View a page containing a search form
3. Open the HTTP API panel in Query Monitor and observe that a request to ES is performed
4. Reload the page a few times and confirm the request from Altis Enhance Search occurs on every page load

This request comes from the `split_large_ep_request()` function that's hooked into the `ep_do_intercept_request` filter. This filter runs when `\ElasticPress\Feature\Autosuggest\generate_search_query()` is called within `\ElasticPress\Feature\Autosuggest\enqueue_scripts()` to enqueue the EP autosuggest JS.

EP Autosuggest adds its own `intercept_search_request()` callback to the `ep_do_intercept_request` filter which performs an HTTP API request but is wrapped in a caching layer.

**_This needs thorough testing_** as I don't know what other impact this change could have.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -1950,7 +1950,7 @@ function custom_search_results_post_type_args( array $args, string $post_type ) 
 function split_large_ep_request( $request, array $query, array $args, int $failures ) {
 
 	if ( empty( $args['body'] ) || $args['method'] !== 'POST' || ! strpos( $query['url'], '/_bulk' ) ) {
-		return wp_remote_request( $query['url'], $args );
+		return $request;
 	}
 
 	// phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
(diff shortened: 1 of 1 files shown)


roborourke:
Gonna see if the current tests pass. I think it'd be good to devise a new test for this change but I was having all sorts of trouble with codeception/selenium and the autosuggest dropdown. The change looks sensible to me. I think we can run with on master at least for a while and do some bulk indexing with lots of posts before we backport it.

roborourke:
So that change to `wp_remote_request()` is there because the default return from that filter is a WP_Error object. I updated the PR to just check for that specifically and create a new remote request object if it is an error.

Also noticed that internally EP is using the intercept request filter by adding and then removing it, so I changed the priority on our one just in case it gets removed by core EP.

roborourke (approved):
Tests are passing so I think this is good to ship now.

PR #409, opened 2022-05-16
Update Elasticsearch versions in v9 docs

johnbillion:
This needs verifying by the cloud team for correctness. Based on https://docs.altis-dxp.com/guides/upgrading/v9/ and https://docs.altis-dxp.com/v10/search/elasticsearch-specifications/.


roborourke (approved):
Quite right, thanks @johnbillion - this was based on EP support at the time I think but we never ran ES 1.x or 2.x at any time!

Issue #410, opened 2022-05-17
Implement proxy endpoint for ElasticPress instant results


roborourke:
The Instant Results feature that landed alongside Altis v11 is not currently fully implemented as I misunderstood what it did.

We need to implement a version of https://github.com/10up/elasticpress-proxy that works for Altis. Notably it _cannot_ use the filesystem other than `/tmp` - and then it must also properly handle running on multiple servers and clearing the cache whenever the site is reindexed.

Acceptance criteria:

- [ ] Functionality of https://github.com/10up/elasticpress-proxy is replicated for Altis
- [ ] Config option for enabling `instant-results` feature
- [ ] Instant Results is switched off by default
- [ ] Docs added for new instant results feature including how to activate it and what it does - namely the search modal overlay. Add a screenshot as well using the twentytwentytwo theme and some demo content & tags
- [ ] Feature works on multisite
- [ ] DB queries can be cached

veselala:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Mercury-6284e0bfb386b00d11981426/issues/humanmade/altis-enhanced-search/410?planning-poker) with ZenHub @robindevitt @roborourke @shadyvb

roborourke:
@humanmade/altis-vulcan drop questions in here or in the product team chat if you have any and I'll answer them.

veselala:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Vulcan-6284e0dfb386b00c83981426/issues/humanmade/altis-enhanced-search/410?planning-poker) with ZenHub @ferschubert-hm @jerico @kovshenin @mikelittle @wisyhambolu

CStukey:
@jerico We refined and estimated during Aug 10 refinement meeting.

CStukey:
@jerico we refined and estimated this ticket during the Aug 10 refinement meeting.

PR #411, opened 2022-05-17
Support for EP instant results custom proxy

roborourke:
Reimplements the custom proxy from https://github.com/10up/elasticpress-proxy

Currently WIP, still to do:

- [ ] Implement request signing for cloud environments
- [ ] Come up with a way to cache the db query results that can be cleared on index/reindex so query only occurs on each container once. Will need to delete cache on reindex if possible

Fixes https://github.com/humanmade/altis-enhanced-search/issues/410

Feature was revealed for v11 so should be backported to v11-branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
 		"files": [
 			"inc/namespace.php",
 			"inc/analysis/namespace.php",
-			"inc/packages/namespace.php"
+			"inc/packages/namespace.php",
+			"inc/instant_results/namespace.php"
 		]
 	},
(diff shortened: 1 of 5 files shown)


roborourke:
cc @stuartshields - it's functional like this on local but does a db query every time right now. Need to fix that and request signing / authentication by changing the curl stuff to use Guzzle directly. I'm away for a week so if you're able to work on it feel free, otherwise it'll be sometime the week after next before it's picked up again.

PR #417, opened 2022-07-12
Change priority on ep_do_intercept_request

### LINKED ISSUE #414: Unnecessary HTTP API request still remaining on pages that enqueue autosuggest
In #405 a change was made to remove an unnecessary HTTP API request to Elasticsearch when the autosuggest JavaScript is enqueued. This change isn't having an effect, presumably because of what @roborourke pointed out:

> So that change to `wp_remote_request()` is there because the default return from that filter is a `WP_Error` object

If the default return from that filter is a `WP_Error` then this request will still fire with the updated logic in #405.

We've deployed 11.0.2 to a client site and we're still seeing this request occurring.

## Steps to reproduce:

1. Enable autosuggest in composer.json
   ```json
   "search": {
   	"autosuggest": true
   },
   ```
2. View a page containing a search form
3. Open the HTTP API panel in Query Monitor and observe that a request to ES is performed
4. Reload the page a few times and confirm the request from Altis Enhance Search occurs on every page load

kovshenin:
In #405 we attempted to get rid of an HTTP request caused by a dummy call to `WP_Query` by EP's Autosuggest feature. However, the request continued to show up as our call on `ep_do_intercept_request` runs before the [intercept callback](https://github.com/humanmade/altis-enhanced-search/blob/bbd46c07c043406006978f2c6cd07c6bf11c7a3d/lib/elasticpress/includes/classes/Feature/Autosuggest/Autosuggest.php#L607) in Autosuggest, which converts the initial `WP_Error` to an HTTP response, either by running `wp_remote_request` or obtaining the results from cache.

This commit sets our `ep_do_intercept_request` to run at a later priority, meaning that if Autosuggest's intercept is set to run,
we will no longer get the default `WP_Error` object, rendering the rest of the logic behind #405 accurate.

Fixes #414

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -202,7 +202,7 @@ function load_elasticpress() {
 	add_action( 'init', __NAMESPACE__ . '\\configure_documents_feature', 1 );
 
 	// Use ElasticPress request intercepting to chunk large requests.
-	// Priority 11 to force these after ElasticPress\Feature\Autosuggest's dummy interceptions
+	// Priority 11 to force these after ElasticPress\Feature\Autosuggest's dummy interceptions.
 	add_filter( 'ep_intercept_remote_request', '__return_true', 11 );
 	add_filter( 'ep_do_intercept_request', __NAMESPACE__ . '\\split_large_ep_request', 11, 4 );
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -203,7 +203,7 @@ function load_elasticpress() {
 
 	// Use ElasticPress request intercepting to chunk large requests.
 	add_filter( 'ep_intercept_remote_request', '__return_true', 11 );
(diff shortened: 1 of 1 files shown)


johnbillion:
Tested on my local and appears to be working as expected.

* The HTTP API request is no longer performed during page generation
* Autosuggest results work as expected
* Autosuggest query is routed to ES

johnbillion (approved):

roborourke:
@kovshenin the backport PRs are typically almost instantaneous, and because the bot user creates them you can approve and merge them yourself

kovshenin:
@roborourke good to know :+1:

kovshenin:
Left some code review on the ElasticPress repo related to my findings here: https://github.com/10up/ElasticPress/issues/2887

PR #419, opened 2022-07-15
Use ElasticPress package and update to v4.2.2.

### LINKED ISSUE #413: Update ElasticPress v4.2
ElasticPress 4.2 is released, so we need to update the package.

Two main changes that we're eyeing for the update:

- Reverting back the change to manually include the files in our repo, and use the composer package directly as we did before the change, as the package release process has been fixed to include assets as it used to. See https://github.com/10up/ElasticPress/pull/2799
- There was a bug that prevented indexing specific sites, where `--url` was ignored, that has been fixed as well. See https://github.com/10up/ElasticPress/issues/2766

Acceptance criteria:

- [ ] ElasticPress packages is used, and its files that were manually added are removed.
- [ ] Basic ElasticPress / Search functionalities are tested, both the manual and the automated parts as per our QA process

robindevitt:
Fixes #413 

ElasticPress packages is used, and its files that were manually added are removed.


roborourke (approved):
Looks good to me, cheers @robindevitt

PR #438, opened 2022-10-19
Only index users once

tfrommen:
The current suggestion for recreating the entire search index will result in the user index being recreated for every single site!

This PR changes the CLI commands to only handle post and term indexes for each site, and then follow up with the user index just once.


roborourke (approved):
Good shout, thanks @tfrommen

PR #486, opened 2023-08-10
update es index command to new sync command

owaincuvelier:


mikelittle (approved):
Looks good. Thanks, Owain. 
One question, would this be better merged into `master` and back ported to `v16-branch`?
Probably a 'tomato', 'tomahto' question.
Regardless, in its current form it needs a 'backport-master` label.

owaincuvelier:
@mikelittle I'm not 100% on the process wrt backporting. I can change the branch if that's what you're saying

mikelittle:
> @mikelittle I'm not 100% on the process wrt backporting. I can change the branch if that's what you're saying

No need to change in this case. 

But generally, new things go in `master` first and thus end up in a new release. Fixes go in the newest branch (or the branch with the bug) and are back ported to master and any other branches as required. Hmmm, I've just talked myself out of my own suggestion 😆 .  As you were. 
Don't forget the label before you merge.

mikelittle:
Except there isn't a `backport-master`, label. But neither is there a  `backport v16-branch` label. Curious!

PR #520, opened 2024-07-05
Fix ElasticPress warning

mikelittle:
Call the CLI command `sync` instead of the deprecated `index` command.


ferschubert-hm (approved):
LGTM

PR #535, opened 2025-02-13
257 document optional containers

mikelittle:
Remove errors when elasticsearch is turned off

When turning off elasticsearch for local server, ensure we don't load elasticpress or try to contact the ES host.

Addresses https://github.com/humanmade/altis-local-server/issues/257


ferschubert-hm (approved):
Syntax review only. Looks good.

PR #542, opened 2025-05-20
Fix local server startup

mikelittle:
Fixes local server startup when root composer.json has no module definitions.


ferschubert-hm (approved):
Tested on my local, proposed changes allow a new altis-local-server env to be provisioned. Approving

PR #575, opened 2026-05-08
Fix handling of subdomains via the proxy

mikelittle:
This fix sets the correct Traefik rules for the Kibana subdomain.

- References PR https://github.com/humanmade/altis-local-server/pull/922


jerico (approved):
Verified to be working locally

PR #596, opened 2026-07-03
Spike: analyzer fix on v24-branch (investigation, do not merge)

mikelittle:
Investigation only — triggers module CI to test whether backporting c37d94970 affects the ElasticSearchCest synonym/stopword failures on EP 4.7. The fix targets EP 5.x stemming, so it's expected to be inert here; this run tells us the baseline (2-fail vs all-fail) on the canonical harness. No reviewers by design.


mikelittle:
Test only - closing now

